PR #13, opened 2018-03-13
Render mobileapp ads

jaiminpanchal27:


src/creative.js
```diff
@@ -120,35 +133,96 @@ function renderCrossDomain(adId, pubUrl) {
...
-    if (bidObject.adm && bidObject.nurl) {
-      ad = bidObject.adm;
-      ad += utils.createTrackPixelHtml(decodeURIComponent(bidObject.nurl));
-      utils.writeAdHtml(ad);
-    } else if (bidObject.adm) {
-      ad = bidObject.adm;
+    let width = (bidObject.width) ? bidObject.width : bidObject.w;
+    let height = (bidObject.height) ? bidObject.height : bidObject.h;
+    if (bidObject.adm) {
+      ad = (!isMobileApp) ? bidObject.adm : constructMarkup(bidObject.adm, width, height);
```
mkendall07: can you swap this logic (don't use !) ? It's confusing to read written this way.

src/creative.js
```diff
@@ -120,35 +133,96 @@ function renderCrossDomain(adId, pubUrl) {
...
+      if (bidObject.nurl) {
+        ad += utils.createTrackPixelHtml(decodeURIComponent(bidObject.nurl));
+      }
       utils.writeAdHtml(ad);
     } else if (bidObject.nurl) {
-      let nurl = bidObject.nurl;
-      utils.writeAdUrl(nurl, bidObject.h, bidObject.w);
+      if(!isMobileApp) {
```
mkendall07: don't need ! here either

src/utils.js
```diff
@@ -61,3 +61,48 @@ export function insertElement(elm, doc, target) {
...
+  });
+  return uuid;
+};
+
+export function loadScript(currentWindow, tagSrc, callback) {
+  let doc = currentWindow.document;
+  let scriptTag = doc.createElement('script');
+  scriptTag.type = 'text/javascript';
+
+  // Execute a callback if necessary
```
mkendall07: this is dead code

src/creative.js
```diff
@@ -120,35 +133,96 @@ function renderCrossDomain(adId, pubUrl) {
...
+        let nurl = bidObject.nurl;
+        utils.writeAdUrl(nurl, width, height);
+      } else {
+        let adhtml = utils.loadScript(window, bidObject.nurl);
+        ad = constructMarkup(adhtml.outerHTML, width, height);
```
mkendall07: interesting approach!

src/creative.js
```diff
@@ -120,35 +133,96 @@ function renderCrossDomain(adId, pubUrl) {
...
-      bidObject = JSON.parse(response);
-    } catch (error) {
-      console.log(`Error parsing response from cache host: ${error}`);
-    }
-    
+/**
+ * Cache request Callback to display creative
+ * @param {Bool} isMobileApp 
+ */
+function handler(isMobileApp) {
```
mkendall07: this could have a more descriptive name

src/creative.js
```diff
@@ -27,20 +27,28 @@ const DEFAULT_CACHE_PATH = '/pbc/v1/cache';
...
+ * Public render ad function to be used in dfp creative setup
  * @param  {object} doc
  * @param  {string} adId
  * @param  {dataObject} dataObject
  */
 pbjs.renderAd = function(doc, adId, dataObject) {
-  if (environment.isAmp(dataObject)) {
-    renderAmpAd(dataObject.cacheHost, dataObject.cachePath, dataObject.uuid);
+  if(environment.isMobileApp(dataObject)) {
+    renderAmpOrMobileAd(dataObject.cacheHost, dataObject.cachePath, dataObject.uuid, true);
```
mkendall07: why pass a flag for isMobile? Just create 2 functions.

jaiminpanchal27:
Updated based on above feedback

src/creative.js
jaiminpanchal27: Updated name to responseCallback

src/creative.js
jaiminpanchal27: Lot of code is common between these two functions so decided to added a flag

PR #14, opened 2018-03-21
Amp resize

jaiminpanchal27:


bretg:
good stuff. To be clear, this isn't just for 1x1, it looks to resize the iframe to the chosen size regardless of the original size.

So it's also useful for multi-size ad units. e.g. 320x50 and 300x50

PR #25, opened 2018-09-17
Tag 1.0 updates

jsnellbaker:
The following changes are included:
- add support for `adServerDomain` param in tag for safeframe environments.  This allows the publisher to set a value for the safeframe domain (previously used hard-coded DFP safeframe domain)
- add support to use DFP `%%PATTERN:TARGETINGMAP%%` macro to dynamically pass in the auction keys to the universal creative rendering function.  The macro provides an object map with each auction key used for that DFP creative; the values for each key are stored in a single array string.
- reworked the tag structure/API to support the above TARGETINGMAP approach and the old approach (for non-DFP) ad-servers.  See example in `dfp-example.html` file for reference.
    - renamed global object `pbjs` to `ucTag` when calling the `renderAd` function

### DIFF
--- a/src/creative.js
+++ b/src/creative.js
@@ -12,7 +12,8 @@
 import * as utils from './utils';
 import * as environment from './environment';
 
-const pbjs = window.pbjs = (window.pbjs || {});
+// const pbjs = window.pbjs = (window.pbjs || {});
+const ucTag = window.ucTag = {};
 const GOOGLE_IFRAME_HOSTNAME = '//tpc.googlesyndication.com';
 const DEFAULT_CACHE_HOST = 'prebid.adnxs.com';
 const DEFAULT_CACHE_PATH = '/pbc/v1/cache';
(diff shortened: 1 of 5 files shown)


src/creative.js
```diff
@@ -29,18 +29,19 @@ const DEFAULT_CACHE_PATH = '/pbc/v1/cache';
...
- * @param  {string} adId
  * @param  {dataObject} dataObject
  */
-pbjs.renderAd = function(doc, adId, dataObject) {
-  if(environment.isMobileApp(dataObject)) {
-    renderAmpOrMobileAd(dataObject.cacheHost, dataObject.cachePath, dataObject.uuid, dataObject.size, true);
-  } else if (environment.isAmp(dataObject)) {
-    renderAmpOrMobileAd(dataObject.cacheHost, dataObject.cachePath, dataObject.uuid, dataObject.size);
+ucTag.renderAd = function(doc, dataObject) {
+  const auctionData = utils.transformAuctionTargetingData(dataObject);
```
jaiminpanchal27: Can you rename auctionData. It store targeting key values so maybe something like `targetingData` `targetingInfo` etc

src/environment.js
```diff
@@ -81,8 +81,8 @@ export function isCrossDomain() {
 
 /**
  * Returns true if envrionment is mobile app
- * @param {Object} dataObject 
+ * @param {String} env
```
jaiminpanchal27: Add @returns here

src/environment.js
```diff
@@ -48,10 +48,10 @@ function isDfp() {
 /**
  * @returns true if there is an AMP context object
  */
-export function isAmp(dataObject) {
+export function isAmp(uuid) {
```
jaiminpanchal27: Add @param here

template/amp/dfp-creative.html
```diff
@@ -1,17 +1,24 @@
...
+  // if not DFP, use these params
+  ucTagData.adId = "%%PATTERN:hb_adid%%";
+  ucTagData.cacheHost = "%%PATTERN:hb_cache_host%%";
+  ucTagData.cachePath = "%%PATTERN:hb_cache_path%%";
+  ucTagData.uuid = "%%PATTERN:hb_cache_id%%";
+  ucTagData.mediaType = "%%PATTERN:hb_format%%";
+  ucTagData.env = "%%PATTERN:hb_env%%";
+  ucTagData.size = "%%PATTERN:hb_size%%";
+
+  // if DFP, use this param
```
jaiminpanchal27: Should we move this up (after line 5) ? Because most of the time adserver will be DFP.

PR #26, opened 2018-09-17
upload test pages for universal creative

jsnellbaker:
This PR includes a number of test pages that were created for Universal Creative.

They cover the following scenarios where universal creative may be used:
- banner in safeframe
- banner in non-safeframe
- outstream in non-safeframe*
- amp (note - uses safeframe)
- mobile-app**

The majority of these pages demonstrate the normal end-to-end workflow, including calling Prebid and DFP to render the test ads.  There is an exception in one case, that's detailed further below.

* - there should be a test page that uses outstream in safeframe, but this is pending and will come later in a subsequent PR
** - due to the heavy environment requirements normally involved for this type of delivery, most of the end-to-end was not included.  The test page enters at the point where prebid-cache is called to grab the ad html.  Here we called a static prebid-cache response that's been included in the project (`cached_mobile_bid.json`) via this PR.  
Note - you need to be hosting the page in a secure environment due to the normal calls to prebid-cache are secure.  If you're running the project locally for testing, simply run `gulp serve --https` and your test page can be loaded securely (after the normal browser warnings).


jsnellbaker:
@Guilemon The test pages have to be hosted some where for it load properly.  

If you want to test - you should be able to run this project locally (ie `gulp serve`) and then load the testpage like: `localhost:9999/testpages/hello_world_amp.html` (or whatever value if you have an alias for localhost).  This should load the testpage in a working environment.

jsnellbaker:
@jaiminpanchal27 I updated the outstream non-sf page and made some additions to the amp page.

I also add the outstream sf page for review.

PR #29, opened 2018-09-24
Unit test setup and code refactor

jaiminpanchal27:
Unit test setup and code refactor.

New gulp commands
- gulp test
- gulp test --watch
- gulp test --browserstack
- gulp test-coverage
- gulp view-coverage


src/utils.js
```diff
@@ -169,4 +169,19 @@ export function transformAuctionTargetingData(dataObject) {
   });
 
   return auctionData;
+}
+
+export function parse(url) {
```
jsnellbaker: Could we use a more descriptive name than `parse`?

gulpfile.js
```diff
@@ -80,3 +83,37 @@ gulp.task('connect', () => {
   });
 });
 
+gulp.task('test', (done) => {
```
jsnellbaker: A few points on this task:

- can you add a comment listing/describing the different known flags  you can pass in the test command?  eg `browserstack`, `https`

- while running the command, it lists out all the packages/files that are being built.  This is a little spammy in the console. Can you see if you can hide that output?  I don't think it's there in prebid.js.

- can you add this task as a dependency step to the `serve` task?

src/renderingManager.js
```diff
@@ -0,0 +1,264 @@
...
+      renderLegacy(doc, targetingData.adId);
+    }
+  };
+
+  /**
+   * Calls prebid.js renderAd function to render ad
+   * @param {Object} doc Document
+   * @param {string} adId Id of creative to render
+   */
+  function renderLegacy(doc, adId) {
```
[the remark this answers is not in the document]
jaiminpanchal27: Updated code to use correct doc.

PR #33, opened 2018-09-28
fix for targeting keys bug

jsnellbaker:
**Bug Summary**
If someone set a property in the `ucTagData` object to an empty string, it would use that value even if it had a proper value in the `ucTagData.targetingMap`.  

For example (contrived in presentation to demonstrate problem):
```
ucTagData.targetingMap.hb_adid = '123abc';
ucTagData.adId = '';
```
This would result in the return object's `auctionData.adId` to be set to `''` instead of `123abc`, which would cause the rendering logic later to fail due to a missing `adId`.

**Cause**
Bad logic in function.

**Fix**
Any keys that are set outside the `targetingMap` that have an empty string are ignored.  This will preserve any previously set keys from the `targetingMap`.  

This change also has a separate benefit in that we're standardizing the appearance of the return object from the `utils.transformAuctionTargetingData` function.  

To illustrate, take the `cachePath` property.  
- In a banner tag for DFP (using `targetingMap`), this key would not be part of the `targetingMap` - so it would be an `undefined` key in the return object.
- In a banner tag for non-DFP (where we set the `ucTagData.cachePath` property), this would be set to an empty string (due to how the token would likely resolve).  In turn we would have an empty string in the return object for the same property.

When the logic in the `renderingManager.js` file's `renderAd` function would have to read this return object - it would have to handle either an `undefined` or an empty string.

With this fix, it will always be `undefined`.

_Note_ - because values are `undefined` instead of empty strings, I set a default value in the argument list for some of the other functions when they are expecting a string type value.  This is to prevent undefined exceptions in certain use-cases.

### DIFF
--- a/src/renderingManager.js
+++ b/src/renderingManager.js
@@ -1,6 +1,6 @@
 import * as utils from './utils';
 
-const GOOGLE_IFRAME_HOSTNAME = '//tpc.googlesyndication.com';
+const GOOGLE_IFRAME_HOSTNAME = 'tpc.googlesyndication.com';
 const DEFAULT_CACHE_HOST = 'prebid.adnxs.com';
 const DEFAULT_CACHE_PATH = '/pbc/v1/cache';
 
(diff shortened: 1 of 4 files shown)


jsnellbaker:
I updated this PR based on the latest changes from the master branch (which already had the cross domain protocol fix).

Please review.

PR #36, opened 2018-10-11
initial version of nativeTrackerManager.js and updates to gulp file

jsnellbaker:
**DO NOT MERGE IMMEDIATELY** see note at bottom...

This PR adds a new script that's meant to complement the native creative template for Prebid.js.

The script is meant to automatically manage the impression/click tracking URLs that are returned with a native bid.  Impression URLs are automatically executed when the script loads, while the clicks are executed when a click event occurs on the creative.  

For reference, below is a copy of the native html snippet that works with this js file (note the CDN path used below is just temporary):
```
<div class="sponsored-post">
  <div class="thumbnail"></div>
  <div class="content">
    <h1><a href="%%CLICK_URL_UNESC%%%%PATTERN:hb_native_linkurl%%" target="_blank" class="pb-click" pbAdId="%%PATTERN:hb_adid%%">%%PATTERN:hb_native_title%%</a></h1>
    <p>%%PATTERN:hb_native_body%%</p>
    <div class="attribution">%%PATTERN:hb_native_brand%%</div>
  </div>
</div>
<script src="https://acdn.adnxs.com/prebid/universal-creative/nativeTrackerManager.js"></script>
```

Additionally - we're making updates to the gulp file to build the file.  Deployment will be automatically managed when we publish to NPM.  The native file will be hosted on jsDelivr in the future.

Note - once the native script file is in it's final state, we will upload a minified version of the file in the `dist` folder as part of it's initial deployment to jsDelivr.  This will be done in a future commit with this PR.


src/nativeTrackerManager.js
```diff
@@ -0,0 +1,79 @@
...
+      message.action = 'click';
+    }
+
+    window.parent.postMessage(JSON.stringify(message), '*');
+  }
+}
+
+// START OF MAIN CODE
+function start() {
+  var adElements = findAdElements(AD_ANCHOR_CLASS_NAME);
```
jaiminpanchal27: Use let/const throughout this file

src/nativeTrackerManager.js
```diff
@@ -0,0 +1,79 @@
...
+    console.warn('Prebid tracking event was missing \'adId\'.  Was adId macro set in the HTML attribute ' + AD_DATA_ADID_ATTRIBUTE + 'on the ad\'s anchor element');
+  } else {
+    var message = { message: 'Prebid Native', adId: adId };
+
+    // fires click trackers when called via link
+    if (action === 'click') {
+      message.action = 'click';
+    }
+
+    window.parent.postMessage(JSON.stringify(message), '*');
```
jaiminpanchal27: Why are we using '*'. Any way to know the targetOrigin

src/nativeTrackerManager.js
```diff
@@ -0,0 +1,79 @@
+/*
+ * Script to handle firing impression and click trackers from native teamplates
```
jaiminpanchal27: Can you convert this to factory pattern just like used in renderingManager and env. It will make writing test cases easier. Also add some unit tests.

src/nativeTrackerManager.js
```diff
@@ -0,0 +1,79 @@
...
+ */
+
+const AD_ANCHOR_CLASS_NAME = 'pb-click';
+const AD_DATA_ADID_ATTRIBUTE = 'pbAdId';
+
+// handle addEventListener gracefully in older browsers
+function addEventHandler(element, event, func) {
+  if (element.addEventListener) {
```
jaiminpanchal27: If we are only supporting IE11+ than we are good to use addEventListener.

jsnellbaker:
@jaiminpanchal27 I made some updates based on your feedback.  Let me know what you think.

PR #38, opened 2018-10-12
E2E testing

jaiminpanchal27:
End to end testing using browserstack and webdriverio.
Command to run : `gulp test --e2e --https`

Known limitation
- To test against your local build, we had to give local path in dfp creative. Redirecting request is not possible in browserstack.
- On localhost domain you do not receive bid, hence you have to have some host defined in hosts file. As of now we are using `test.localhost` defined in local host file as well as `DFP creative`.

Above limitation can be removed by moving files to internet accessible box but we have monitoring setup and purpose of e2e testing was more to catch the bugs while developing.

Todos:
- Gulp 4 has good features to run tasks in series and parallel. Add gulp 4
- Amp and Banner Safeframe is not tested since they are both rendered in cross domain iframe. Find some workaround to test these two.


gulpfile.js
```diff
@@ -16,10 +16,12 @@ var rename = require('gulp-rename');
 var KarmaServer = require('karma').Server;
 var opens = require('open');
 var karmaConfMaker = require('./karma.conf.maker');
+var execa       = require('execa');
```
mkendall07: prefer const + spacing.

test/e2e/specs/hello_world_banner_non_sf.spec.js
```diff
@@ -0,0 +1,16 @@
+const assert = require('assert');
+
+describe.only('banner', function() {
+    it('should load the creative', function() {
+				browser
```
mkendall07: spacing looks weird here.

PR #45, opened 2018-12-03
Adding a query param to limit the /cookie_sync pixel count

### LINKED ISSUE #44: Add query param to limit cookie syncs
**Is your feature request related to a problem? Please describe.**

As more bidders get added to Prebid Server, it returns more and more usersync pixels. This is starting to bog down the performance of pages.

This is mainly targeted at AMP pages. They load the cookie sync script from the CDN: http://prebid.org/dev-docs/show-prebid-ads-on-amp-pages.html#user-sync

**Describe the solution you'd like**

Make the cookie sync script check the query params for a "sync limit". For example, `https://acdn.adnxs.com/prebid/amp/user-sync/load-cookie.html?max_sync_count=5`. Make sure the script doesn't fire any more syncs than that.

**Describe alternatives you've considered**

none... AMP doesn't give us much flexibility here. I can imagine some other future features (e.g. publishers choosing which syncs to prioritize)... but nothing that would conflict with this, from an API standpoint.

dbemiller:
Fixes #44.

This also replaces gulp-connect with gulp-webserver because of that same `flatmap-stream` vulnerability... I had to do this to be able to test it.


mkendall07:
nice thanks for the PR! Does it make sense to randomize the order of the syncs before we truncate it for MAX? Thinking here that the same pixels will be fired over and over and some bidders may never see the pixels fired because they are later in the list...

dbemiller:
PBS has a related PR: https://github.com/prebid/prebid-server/pull/759... we should probably wait until that's released and then use the server-side limit.

Server-side limiting is better for a few reasons (less bandwidth, easier to roll out bugfixes + smarter sync logic, etc)

PR #49, opened 2018-12-13
Run the cookie syncs in series

### LINKED ISSUE #43: Usersync pixels should be run in series
**Describe the bug**

This project calls `/cookie_sync`, but then runs all the returned syncs in parallel. See: https://github.com/prebid/prebid-universal-creative/blob/master/src/cookieSync.js#L24

If more than one call to `/setuid` is made at the same time, then the Cookie after all syncs will be missing some values. For more details, see the additional context.

**To Reproduce**

Since this is a concurrency issue, it cannot be reproduced reliably. It has a chance of happening on any page that uses `src/cookieSync.js` though.

**Expected behavior**

After all the `/setuid` calls have completed, the final cookie should contain all the new syncs that were just done.

**Additional context**

Prebid Server's `/setuid` endpoint behaves like so:

1. Unpack the Cookie sent in the request, to get existing syncs.
2. Update the data inside the Cookie with the new Bidder/UID
3. Respond with a `Set-Cookie` header with the updated data.

In practice, this means that it can't handle parallel requests from the same page. If a second `/setuid` . call occurs before the first one returns, then the same (old) cookie will get sent with both. Afterwards, the cookie will only have the sync for whichever `/setuid` call returned _last_.

dbemiller:
Fixes #43. Also present in https://github.com/prebid/Prebid.js/issues/3265.


src/cookieSync.js
```diff
@@ -14,33 +14,49 @@ const MAX_SYNC_COUNT = sanitizeSyncCount(parseInt(parseQueryParam('max_sync_coun
...
     console.log(`User sync type "${type}" not supported for bidder: "${bidder}"`);
+    done();
   }
 }
 
-function triggerPixel(url) {
+function triggerPixel(url, done) {
   const img = new Image();
+  img.addEventListener('load', done);
+  img.addEventListener('error', done);
```
mkendall07: good call on error!

src/cookieSync.js
```diff
@@ -14,33 +14,49 @@ const MAX_SYNC_COUNT = sanitizeSyncCount(parseInt(parseQueryParam('max_sync_coun
...
   img.src = url;
 }
 
+function doAllSyncs(bidders) {
+  if (bidders.length === 0) {
+    return;
+  }
+
+  const thisSync = bidders[bidders.length - 1];
+  bidders = bidders.slice(0, bidders.length - 1);
```
jsnellbaker: As a small question - wouldn't you be able to use `const thisSync = bidders.pop()` instead of using both of these statements?

dbemiller: yes! I didn't know that was a function in JS. Definitely better that way, though.

PR #50, opened 2018-12-17
Rad 2607 update cookie sync

deekshithraop:


deekshithraop:
@jsnellbaker conflicts have been resolved

src/cookieSync.js
```diff
@@ -9,6 +9,8 @@
 
 const ENDPOINT = 'https://prebid.adnxs.com/pbs/v1/cookie_sync';
 const MAX_SYNC_COUNT = sanitizeSyncCount(parseInt(parseQueryParam('max_sync_count', window.location.search), 10));
+const GDPR = sanitizeGdpr(parseInt(parseQueryParam('gdpr', window.location.search)));
```
jaiminpanchal27: pass radix as 2nd argument on parseInt. Javascript will assume radix if it is not passed. More info here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt

deekshithraop: Added base 10 as radix.

PR #55, opened 2019-02-19
support iframe type syncs

mkendall07:
adds iframe support. Small refactor. Should probably add tests....


src/cookieSync.js
```diff
@@ -28,15 +30,26 @@ function doBidderSync(type, url, bidder, done) {
...
+    triggerIframeLoad(url, bidder, done);
   } else {
     console.log(`User sync type "${type}" not supported for bidder: "${bidder}"`);
     done();
   }
 }
 
+function triggerIframeLoad(url, bidder, done) {
+  let iframe = domHelper.getEmptyIframe(0, 0);
+  iframe.id = `sync_${bidder}_${Date.now()}`;
```
deekshithraop: shouldn't these 2 statements be after the the `if` condition check for url

mkendall07: yeah good point

PR #58, opened 2019-02-26
add appnexus and extra param support to endpoint config

idettman:
Added endpoint configuration for `appnexus`. 

Also, we need to pass the Rubicon account ID to satisfy our GDPR rules. This is added as part of the POST body. 

`load-cookie.html?endpoint=rubicon&args=account:1001,other:value`


bretg:
Tested this and found that the limit value was a large number by default. Big enough that it caused Prebid Server to choke.

Isaac and I figured out that specifying max_sync_count was a workaround, but that value needs a default.

bretg:
We also tried different combinations of parameters and found that it automatically adds quotes around both the attribute and the value.

I've taken a cut at updating the prebid.org docs -- https://github.com/prebid/prebid.github.io/pull/1312

bretg:
Much better - thanks Isaac. @mkendall07 - over to you.

bretg:
Docs PR https://github.com/prebid/prebid.github.io/pull/1312

mkendall07:
looks ok to me but should we have a larger default value than 5? seems a bit small to me.

bretg:
Great - how is the universal creative deployed to jsdelivr?

mkendall07:
to deploy we use NPM. https://github.com/prebid/prebid-universal-creative#build-for-production

You'll need create an NPM user and then I can grant you access to the prebid repos.

PR #59, opened 2019-04-04
Buyer user ids script

jaiminpanchal27:
This script gets the buyer user ids from prebid server getuids endpoint.
Use case for now: In Prebid server long form support, SSAI vendor needs to get buyer user id to send it in prebid server request payload. 

It will be available via jsDelivr. Path to be shared soon
Usage. 
```
<script type="text/javascript" src="<uid url>"></script>
<script type="text/javascript">
  window.pbs.getBuyerUids(function(err, uids) { 
    // Use uids here
  })
</script>
```

### DIFF
--- a/src/ssp-userids/uid.js
+++ b/src/ssp-userids/uid.js
@@ -22,7 +22,7 @@ const REFRESH_IN_DAYS = 14;
  */
 export function getBuyerUids(callback) {
   let uidData = commons.getDataFromLocalStorage(STORAGE_KEY);
-  if (hasValidData(uidData)) {
+  if (hasInvalidData(uidData)) {
     loadData(callback);
   } else {
     try {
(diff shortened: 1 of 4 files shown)


src/ssp-userids/uid.js
```diff
@@ -0,0 +1,74 @@
...
+ * Check whether buyer id data is not null and is not expired
+ * @param {Object} uidData 
+ * @returns {boolean}
+ */
+function hasValidData(uidData) {
+  return !uidData || commons.timestamp() < uidData.lastUpdated + REFRESH_IN_DAYS * 24 * 60 * 60 * 1000
+}
+
+function noop() {}
+loadData(noop);
```
jsnellbaker: Can you clarify why this call is included in the file?  

My assumption is that it's meant to automatically call for the user's data and store it when the script loads.  

If so, are there any plans to allow someone to call the `loadData` function externally of the script so they can update the data of the user on their discretion?

If not, is this call just a temporary thing?

jsnellbaker: Spoke separately; all good at this time.

PR #62, opened 2019-05-16
replace ${AUCTION_PRICE} in the bidObject.adm when rendering in amp

JonGoSonobi:
Fixes https://github.com/prebid/prebid-universal-creative/issues/61

### DIFF
--- a/src/renderingManager.js
+++ b/src/renderingManager.js
@@ -182,7 +182,7 @@ export function newRenderingManager(win, environment) {
       let width = (bidObject.width) ? bidObject.width : bidObject.w;
       let height = (bidObject.height) ? bidObject.height : bidObject.h;
       if (bidObject.adm) {
-        if(hbPb) { // replace ${AUCTION_PRICE} macro with the clear price.
+        if(hbPb) { // replace ${AUCTION_PRICE} macro with the hb_pb.
           bidObject.adm = bidObject.adm.replace('${AUCTION_PRICE}', hbPb);
         } else {
           /*
--- a/src/renderingManager.js
+++ b/src/renderingManager.js
@@ -29,9 +29,9 @@ export function newRenderingManager(win, environment) {
   let renderAd = function(doc, dataObject) {
     const targetingData = utils.transformAuctionTargetingData(dataObject);
     if(environment.isMobileApp(targetingData.env)) {
-      renderAmpOrMobileAd(targetingData.cacheHost, targetingData.cachePath, targetingData.uuid, targetingData.size, true);
+      renderAmpOrMobileAd(targetingData.cacheHost, targetingData.cachePath, targetingData.uuid, targetingData.size, targetingData.clearPrice, true);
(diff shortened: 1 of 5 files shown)


JonGoSonobi:
@jaiminpanchal27 @jsnellbaker any updates for this?

JonGoSonobi:
@mkendall07 I map the `hb_pb` from the ucTagData.targetingMap to `clearPrice` check out src/utils.js line 109. I added tests to show this is working in the changes with utils_spec.js. And from what I can tell the template already includes the targetingMap.

jsnellbaker:
While the targetingMap macro would automatically include the hb_pb targeting key and should be read through the changes put together here, there are other formats of the Universal Creative where the variables for the targeting keys are defined specifically and passed into the UC `renderAd` call.  

Those templates would need to be updated in order to populate the `hb_pb` and subsequently the `clearPrice`.  A docs PR should be put together to update the various UC templates that don't rely on targetingMap.

eg the template on http://prebid.org/adops/send-all-bids-adops.html would need to be updated like the following:
```
<script src = "https://cdn.jsdelivr.net/npm/prebid-universal-creative@latest/dist/creative.js"></script>
<script>
  var ucTagData = {};
  ucTagData.adServerDomain = "";
  ucTagData.pubUrl = "%%PATTERN:url%%";
  ucTagData.adId = "%%PATTERN:hb_adid_BIDDERCODE%%";
  ucTagData.cacheHost = "%%PATTERN:hb_cache_host_BIDDERCODE%%";
  ucTagData.cachePath = "%%PATTERN:hb_cache_path_BIDDERCODE%%";
  ucTagData.uuid = "%%PATTERN:hb_cache_id_BIDDERCODE%%";
  ucTagData.mediaType = "%%PATTERN:hb_format_BIDDERCODE%%";
  ucTagData.env = "%%PATTERN:hb_env%%";
  ucTagData.size = "%%PATTERN:hb_size_BIDDERCODE%%";
  ucTagData.clearPrice = "%%PATTERN:hb_pb_BIDDERCODE%%";
  try {
    ucTag.renderAd(document, ucTagData);
  } catch (e) {
    console.log(e);
  }
</script>
```

JonGoSonobi:
Any reservations about naming it `clearPrice` before I go ahead and find all the places in the docs that have this template and add the new key?

mkendall07:
I would actually prefer leaving it as `hb_pb` because that's what it is. Naming it `clearPrice` gives the impression that it's an exact price, but it's actually rounded (via whatever rounding rules have been defined by the publisher).

JonGoSonobi:
@mkendall07 Changes to the Docs are up for PR https://github.com/prebid/prebid.github.io/pull/1329
I also renamed `clearPrice` to `hbPb`.

bretg:
I just merged https://github.com/prebid/prebid-universal-creative/pull/58 - combining the load-cookie changes with this one into one release to NPM.

bretg:
@mkendall07 - are we good here?

PR #63, opened 2019-06-10
🚀 Add Denakop prebid-server endpoint

gurodrigues:
## Type of change
- [ X ] New endpoint

## Description of change
- It was necessary to add Denakop prebid server URL into `cookieSync.js` file

## Other information
- Contact email: gustavo@denakop.com


bretg:
Sorry, but you're going to need to be a prebid.org member before can add your server this way.

You're free to fork this repo and distribute the resulting code.

And honestly, putting servers in load-cookie.html is fairly pointless anyhow because jsdelivr has security settings that prevent AMP iframes from working with it anyhow. So each company has to host it themselves.

PR #67, opened 2019-07-09
Fix e2e test

jaiminpanchal27:
In ssl mode, dfp has added extra div wrapper to load the creative.


jaiminpanchal27:
Closing as it fails in IE 11 & Edge. Needs more investigation

PR #69, opened 2019-07-24
Fix e2e tests

Fawke:
## Type of change

- [X] Build related

## Description of change

Fixed all failing e2e tests. To test, try running this command, `gulp test --e2e --https`

There's just one issue, when running this command, Browserstack is randomly picking a OS version, instead of the one specified in `browsers.json` file.

### DIFF
--- a/testpages/hello_world_banner_non-sf.html
+++ b/testpages/hello_world_banner_non-sf.html
@@ -52,11 +52,6 @@
         });
       });
     }
-
-    // setTimeout(function () {
-    //   sendAdserverRequest();
-    // }, PREBID_TIMEOUT);
-
   </script>
 
   <script>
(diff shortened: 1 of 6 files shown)


Fawke:
Update - the command `gulp test --e2e --https` is working as expected, since we're overriding the `os_version` key in `wdio.conf.js`.

Browserstack is also correctly loading the right versions of the browsers if we run the command, `gulp test --browserstack`.

testpages/hello_world_banner_non-sf.html
```diff
@@ -53,9 +53,9 @@
       });
     }
 
-    setTimeout(function () {
-      sendAdserverRequest();
-    }, PREBID_TIMEOUT);
+    // setTimeout(function () {
```
jaiminpanchal27: Please remove the commented code here and other places

Fawke: @jaiminpanchal27 I have removed the commented lines.

jsnellbaker:
Closing this PR in favor of #72 as that PR appears to already include all these changes anyway.

PR #71, opened 2019-07-24
Fix for so that the impressionTrackers and jsTrackers are called once per adElement

jsnellbaker:
This fix corrects an issue that occurs when multiple native placements use the `native-trk.js` script.

The previous code only grabbed the first adElement and executed the imp/js trackers for that ad.  This fix will detect all available elements and execute them appropriately.


jsnellbaker:
@idettman @jaiminpanchal27 Could you take a look into this PR?

Issue #70, opened 2019-07-24
Support win notifications


bretg:
Prebid Server is accepting events as outlined in https://github.com/prebid/prebid-server/issues/800. It's also generating event ad targeting as `hb_winurl` and 'hb_bidid', e.g.

```
                    "ext": {
                        "prebid": {
                            "type": "banner",
                            "targeting": {
                                "hb_winurl": ""hb_winurl":"https%3A%2F%2Fprebid-server.rubiconproject.com%2Fevent%3Ft%3Dwin%26b%3DBIDID%26a%3D1001%26f%3Di",
                                "hb_bidid_rubicon": "4b61fbb6-d864-4a36-b3dc-ef860d104601",
                                "hb_size_rubicon": "300x250",
                                "hb_cache_id": "731bd2a2-7f84-4080-b499-550a45d4fb9e",
                                "hb_cache_path_rubicon": "/cache",
                                "hb_cache_host_rubicon": "prebid-cache-us-west.rubiconproject.com",
                                "hb_pb": "1.20",
                                "hb_pb_rubicon": "1.20",
                                "hb_cache_id_rubicon": "731bd2a2-7f84-4080-b499-550a45d4fb9e",
                                "hb_cache_path": "/cache",
                                "hb_size": "300x250",
                                "hb_bidder": "rubicon",
                                "hb_bidder_rubicon": "rubicon",
                                "hb_cache_host": "prebid-cache-us-west.rubiconproject.com"
                            },
```
To reduce the footprint of the ad targeting variables the proposal here is to send the whole URL only once, then for each bidder supply the BIDID. So the first two targeting variables combine.

So now it's time to have the universal creative look for the existence of `hb_winurl` and `hb_bidid_BIDDECODE`, hitting the endpoint when appropriate.

The proposed ad server representation is to add a couple of new entries: `ucTagData.winurl` and `ucTagData.winbidid`:
```
<script src = "https://cdn.jsdelivr.net/npm/prebid-universal-creative@latest/dist/creative.js"></script>
<script>
  var ucTagData = {};
  ucTagData.adServerDomain = "";
  ucTagData.pubUrl = "%%PATTERN:url%%";
  ucTagData.adId = "%%PATTERN:hb_adid_BIDDERCODE%%";
  ucTagData.cacheHost = "%%PATTERN:hb_cache_host_BIDDERCODE%%";
  ucTagData.cachePath = "%%PATTERN:hb_cache_path_BIDDERCODE%%";
[...]

PR #75, opened 2019-09-09
add winurl support

idettman:
# Add support for win url

## Description
Prebid Server is being updated to create hb_winurl targeting variables that will be used to track events for mobile apps, AMP, and PG. To make this work we need to update the Prebid Universal Creative.

Updated `renderAd()` to:
* Resolve the 'BIDID' macro in `winurl` with the value of `winbidid`
* If a problem with this process but `hb_winurl` exists, `log` a warning and skip
* Add the URL to an invisible image similar to the **PBJS** `triggerpixel` function in [Prebid.js/src/utils.js](https://github.com/prebid/Prebid.js/blob/master/src/utils.js)

## Related
[Issue #70](https://github.com/prebid/prebid-universal-creative/issues/70)


bretg:
Confirmed that Isaac is ready for review here -- @harpere

test/spec/renderingManager_spec.js
```diff
@@ -89,15 +89,17 @@ describe('renderingManager', function() {
...
     it('should render mobile app creative', function() {
       const renderObject = newRenderingManager(mockWin, env);
       let ucTagData = {
         cacheHost: 'example.com',
         cachePath: '/path',
         uuid: '123',
-        size: '300x250'
+        size: '300x250',
+        winurl: 'https://prebid-server.rubiconproject.com/event?t=win&b=BIDID&f=i',
+        bidid: 'AAAA-BBBB-CCCC-DDDD'
```
harpere: shouldn't this be `winbidid`?

test/spec/renderingManager_spec.js
```diff
@@ -117,7 +119,9 @@ describe('renderingManager', function() {
       const renderObject = newRenderingManager(mockWin, env);
       let ucTagData = {
         uuid: '123',
-        size: '300x250'
+        size: '300x250',
+        winurl: 'https://prebid-server.rubiconproject.com/event?t=win&b=BIDID&f=i',
+        bidid: 'AAAA-BBBB-CCCC-DDDD'
```
harpere: same, shouldn't this be `winbidid`?

test/spec/renderingManager_spec.js
```diff
@@ -169,7 +173,9 @@ describe('renderingManager', function() {
         cachePath: '/path',
         uuid: '123',
         size: '300x250',
-        hbPb: '10.00'
+        hbPb: '10.00',
+        winurl: 'https://prebid-server.rubiconproject.com/event?t=win&b=BIDID&f=i',
+        bidid: 'AAAA-BBBB-CCCC-DDDD'
```
harpere: same, shouldn't this be `winbidid`?

test/spec/renderingManager_spec.js
```diff
@@ -211,12 +217,14 @@ describe('renderingManager', function() {
...
         isCrossDomain: () => true
-      }
+      };
       const renderObject = newRenderingManager(mockWin, env);
       let ucTagData = {
         adId: '123',
         adServerDomain: 'mypub.com',
         pubUrl: 'http://example.com',
+        winurl: 'https://prebid-server.rubiconproject.com/event?t=win&b=BIDID&f=i',
+        bidid: 'AAAA-BBBB-CCCC-DDDD'
```
harpere: again, shouldn't this be `winbidid`?

PR #80, opened 2019-10-02
decoding winurl

bretg:
Did some testing on https://github.com/prebid/prebid-universal-creative/pull/75 and found that we're going to need to encode the hb_winurl or GAM won't pass it through. So adding a level of unencoding.


bretg:
I tested this locally with a stubbed Prebid Server response that had been encoded. Saw the /event URL go out.

PR #81, opened 2019-10-07
fix gulp commands to test and view code coverage

Fawke:
This PR fixes the commands, `gulp test-coverage` & `gulp view-coverage` which were throwing error previously.

As a result of upgrading few of the packages to fix the above commands, node version also had to be upgraded to *8.9.0* in order to support some of the newly installed packages which were using the spread operator.


.babelrc
```diff
@@ -1,13 +1,11 @@
...
 		"transform-es3-property-literals",
-		"transform-es3-member-expression-literals"
-	],
-	"env": {
-    "test": {
-      "plugins": [ "istanbul" ]
-    }
-  }
+		"transform-es3-member-expression-literals",
+		"@babel/plugin-transform-modules-commonjs"
```
jaiminpanchal27: Why do we need `@babel/plugin-transform-modules-commonjs` ?

Fawke: @jaiminpanchal27 We updated the version of `babel-preset-env`, the plugin to transform es6 modules to commonjs modules no longer comes bundled with the new version. We need to install it separately. Since, some of our files are using `import/export` syntax need this plugin or else, the build fails with the error, `exports is not defined.`

jaiminpanchal27:
@Fawke Can you try this https://www.npmjs.com/package/gulp-webserver as alternative to gulp-connect

Fawke:
Closing this for this https://github.com/prebid/prebid-universal-creative/pull/84. This mentioned PR is an offshoot of this PR, so it already has these changes, plus, it also extra changes of refactoring the gulpfile and fixing e2e test cases.

PR #82, opened 2019-10-30
Safeframe fix

benjaminclot:
In some cases I see SafeFrames not being resized properly by GAM. The container DIV is the right size, the Prebid frame is the right size but the SafeFrame in between stays stuck at 1x1. Digging a bit into the code I saw that the cross-domain case doesn't call the `resizeIframe()` method -- maybe because GAM seems to have its own size detection feature... that doesn't always work. So why not use the SafeFrame API and fix that?
Also, calling the `resize()` method right after registering it seems to fix another bug(?) where only registering doesn't always trigger a resize.

The Prebid documentation should also indicate that `allowPushExpansion` has to be `true` for `googletag.SafeFrameConfig` (at least in GAM).

### DIFF
--- a/src/renderingManager.js
+++ b/src/renderingManager.js
@@ -118,11 +118,13 @@ export function newRenderingManager(win, environment) {
         if (adObject.mediaType === 'video') {
           console.log('Error trying to write ad.');
         } else if (ad) {
-          const iframe =  domHelper.getEmptyIframe(adObject.height, adObject.width);
+          const iframe =  domHelper.getEmptyIframe(height, width);
           body.appendChild(iframe);
           iframe.contentDocument.open();
           iframe.contentDocument.write(ad);
           iframe.contentDocument.close();
+          
+          resizeIframe(width, height);
         } else if (url) {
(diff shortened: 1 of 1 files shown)


benjaminclot:
I remembered that resizing the safeFrame is the job of `resizeRemoteCreative()` (https://github.com/prebid/Prebid.js/blob/master/src/secureCreatives.js). It seems to be doing its job but sometimes I don't know what happens, the safeFrame stays at 1x1.

AmerSharif:
Hi. Any update here? We had to deaktive universal creative lineitems in GAM and use the old type until this is fixed.

sumit116:
Hi @benjaminclot , I tested with some example pages but could not reproduce the bug as you mentioned in the description. So, can you share the publisher page where this issue exists? Also, can you resolve the code conflicts?

benjaminclot:
Hi @sumit116 I think you can forget about this PR. I can't reproduce it and anyway I added a safety measure that resizes the iframe on the `bidWon` event.

PR #83, opened 2019-11-13
Mobile native rendering

jaiminpanchal27:
This PR adds support to render native assets in Mobile. Primary use case is when used with Prebid mobile and Prebid server

Docs PR in progress.


src/nativeAssetManager.js
```diff
@@ -27,10 +29,134 @@ const NATIVE_KEYS = {
...
+      assets,
+      clicktrackers,
+      'imptrackers' : adMarkup.imptrackers
+    }
+  }
+
+  function requestAssetsFromCache(tagData) {
+    let ajaxCallback = function(response) {
+      let bidResponse = parseResponse(response);
+      if (bidResponse.adm) {
```
harpere: `bidResponse` will be undefined if the JSON.parse(response) fails, in which case this line fails.

src/nativeAssetManager.js
```diff
@@ -27,10 +29,134 @@ const NATIVE_KEYS = {
...
+        let markup = parseResponse(bidResponse.adm);
+        if (markup.assets) {
```
harpere: same as above, `markup` could be undefined

jaiminpanchal27:
@harpere Thanks.  Updated the code

PR #86, opened 2019-11-19
Use a global regex replace instead of string replace on native placeholders

beth92:
Native asset placeholders are currently replaced using a simple string replace. This replaces the first occurrence of the native asset placeholder in the native template HTML, but does not cover the case where the template contains multiple usages of the placeholder. This can occur because a native template may contain multiple clickable elements. For example, the native image and also the call to action may _both_ link to the click-through URL. 
Currently, the result of running the native tracker script is that the first clickable element will correctly resolve to a valid click-through URL, while any additional links still contain the placeholder `hb_native_linkurl:<ad_id>` and clicking the element results in a blank page / 400 response.
Testing this change locally, I am able to get all clickable elements to direct to a valid URL with no adverse effects, as far as I can tell.

### DIFF
--- a/src/nativeAssetManager.js
+++ b/src/nativeAssetManager.js
@@ -245,7 +245,7 @@ export function newNativeAssetManager(win) {
 
     (assets || []).forEach(asset => {
       const searchString = (adId) ? `${NATIVE_KEYS[asset.key]}:${adId}` : `${NATIVE_KEYS[asset.key]}`;
-      const searchStringRegex = new RegExp(searchString, "g");
+      const searchStringRegex = new RegExp(searchString, 'g');
       html = html.replace(searchStringRegex, asset.value);
     });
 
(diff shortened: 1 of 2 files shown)


src/nativeAssetManager.js
```diff
@@ -244,8 +244,9 @@ export function newNativeAssetManager(win) {
     let html = document;
 
     (assets || []).forEach(asset => {
-      const searchString = (adId) ? `${NATIVE_KEYS[asset.key]}:${adId}` : `${NATIVE_KEYS[asset.key]}`
-      html = html.replace(searchString, asset.value);
+      const searchString = (adId) ? `${NATIVE_KEYS[asset.key]}:${adId}` : `${NATIVE_KEYS[asset.key]}`;
+      const searchStringRegex = new RegExp(searchString, "g");
```
idettman: Could you change the double quotes surrounding `g` to single quotes, the reset of the change is good.

benjaminclot:
@idettman @jaiminpanchal27 Why has this not been pushed to jsdelivr ? Last release is from way back in november 😞

PR #91, opened 2020-01-16
add logic to detect prebid.js global

jsnellbaker:
This is a change to fix an issue reported in Prebid.js (https://github.com/prebid/Prebid.js/issues/4682).

This new logic tries to find the prebid.js global object by climbing the available windows.  If the prebid object isn't found - we're assuming it's blocked by a cross domain iframe.  This will help in situations where prebid.js is hosted inside a cross-domain iframe and the creative.js file can normally directly access prebid.js to render the bid.

We're preserving the old `isCrossDomain` logic for the amp use-case, to avoid the scope of impact.

Also cleaned up some of the formatting.

### DIFF
--- a/src/environment.js
+++ b/src/environment.js
@@ -34,31 +34,48 @@ export function newEnvironment(win) {
     * @returns true if the environment is a Cross Domain
     */
   function isCrossDomain() {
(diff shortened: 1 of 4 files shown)


GeneGenie:
UP =D

PR #95, opened 2020-03-10
Support win URL in cached creatives

idettman:
# Support win URL in cached creatives

## Description

[Related to Issue 93](https://github.com/prebid/prebid-universal-creative/issues/93) 

When Prebid Universal Creative reads from Prebid Cache, we need to have it check for the existence of the wurl parameter. If it exists, hit it.

Note that PBS can respond to event URLs in two ways:
1. blank pixel (if the f=i parameter is specified)
2. HTTP 204 (if the f=b parameter is specified)
PUC may hit the pixel immediately, but in a way that doesn't create console errors or affect screen layout.

Note: that PUC should still support looking for hb_winurl during a transition period.


bretg:
We found an issue in testing that's really a requirements issue. Updated the original issue:
```
In order to avoid tracking things twice, we also need to remove support for hb_winurl in PUC.
This is a breaking change, any entities using hb_winurl need to be aware of this.
I believe Rubicon's the only one using it now, but will update the release notes.
```
Sorry for not specifying this correctly to start with.

src/renderingManager.js
```diff
@@ -39,27 +39,6 @@ export function newRenderingManager(win, environment) {
     } else {
       renderLegacy(doc, targetingData.adId);
     }
-    
-    // check for winurl and replace BIDID token with value if it exists
-    if (targetingData.winurl && targetingData.winbidid) {
```
jaiminpanchal27, line 44: Is this logic not needed anymore ? @SyntaxNode @hhhjort Is this used anywhere in Perbid server ?

hhhjort: This looks like this is firing off a win notification if a win URL is supplied? PBS doesn't touch win notifications, it is not in the loop to know if a bid wins the final auction. So I cannot say if overall if this is needed, we have no metrics on how often such a pixel may be supplied with a bid.

bretg: I made the assumption that PBS-Go hasn't started implementing events yet. If that's true, then there's no backwards compatibility issue. This is essentially changing the design we had done in PBS-Java because it ran into problem in the real world.

PR #103, opened 2020-04-30
native tracker manager click tracker not firing

wqi1972:
Refactor code inside `attachClickListeners`, so that it read adId from the `<a>` tag element instead of read it from `event.target`.
Code read from `event.target` has trouble when there is inner tag inside `<a>` tag, for example:
```<a href="%%CLICK_URL_UNESC%%%%PATTERN:hb_native_linkurl%%" target="_blank" class="pb-click" pbAdId="%%PATTERN:hb_adid%%"><p>%%PATTERN:hb_native_title%%<p></a>```
when click event fires, the `<p>` tag is the event.target, not `<a>` tag, in that case, original code inside function readAdIdFromEvent won't work.

Adjusted test case for it as well.

### DIFF
--- a/src/nativeTrackerManager.js
+++ b/src/nativeTrackerManager.js
@@ -30,19 +30,7 @@ export function newNativeTrackerManager(win) {
     return adId || '';
   }
 
-  function readAdIdFromEvent(event) {
-    let adId =
-      event &&
-      event.target &&
-      event.target.attributes &&
-      event.target.attributes[AD_DATA_ADID_ATTRIBUTE] &&
-      event.target.attributes[AD_DATA_ADID_ATTRIBUTE].value;
-
-    return adId || '';
-  }
-
-  function loadClickTrackers(event) {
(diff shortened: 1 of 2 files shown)


bretg:
@wqi1972 - is this PR needed even if we implement the improvements to native as noted in https://github.com/prebid/Prebid.js/issues/5149?

wqi1972:
@bretg we won't need this PR after prebid/Prebid.js#5149. Inside prebid/Prebid.js#5149 it will fundamentally solve this problem, since new implement will pass adId as parameter, there will be no need to read adId from html element anymore.

If we have some Universal creative release before prebid/Prebid.js#5149 get implemented, I suggest we merge this PR into it, since this bug already impacts some publishers (https://github.com/prebid/prebid-universal-creative/issues/88). If we don't, then we can close this PR and wait for prebid/Prebid.js#5149 solve it fundamentally.

PR #104, opened 2020-05-21
not relying on hard coded google iframe domain

harpere:
to address the problem in https://github.com/prebid/Prebid.js/issues/5026, no longer relying on the hard coded google safeframe domain.

### DIFF
--- a/src/renderingManager.js
+++ b/src/renderingManager.js
@@ -67,7 +67,7 @@ export function newRenderingManager(win, environment) {
    * @param {string} pubUrl Url of publisher page
    */
   function renderCrossDomain(adId, pubAdServerDomain = '', pubUrl) {
-    let windowLocation = window.location;
+    let windowLocation = win.location;
     let parsedUrl = utils.parseUrl(pubUrl);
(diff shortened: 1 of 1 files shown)


src/renderingManager.js
```diff
@@ -71,7 +70,7 @@ export function newRenderingManager(win, environment) {
     let windowLocation = window.location;
     let parsedUrl = utils.parseUrl(pubUrl);
     let publisherDomain = parsedUrl.protocol + '://' + parsedUrl.host;
-    let adServerDomain = pubAdServerDomain || GOOGLE_IFRAME_HOSTNAME;
+    let adServerDomain = pubAdServerDomain || win.location.hostname;
```
wqi1972, line 73: Any particular reason why inside some function, sometime uses `window`, sometime uses `win`?

harpere: no reason.  updated to make it consistent.

bretg:
@wqi1972 - look better to you?

wqi1972:
LGTM

PR #106, opened 2020-06-22
PUC update to include Native Phase 2 spec:

mmoschovas:
1) Added renderNativeAd function to nativeTrackerManager.js
2) renderNativeAd checks for existence of adId and performs the following:
   - Checks for rendererUrl and preloads if exists
   - Uses prexisting loadAssets function with new callback
   - Fires Impression tracker and loads click tracker
3) LoadAssets update:
   - Checks for existence of assetsToReplace and adds each asset to placeholder
   - If no placeholders exist, checks for requestAllAssets flag to fire new requestAllAssets function
4) requestAllAssets will postmessage Prebid using the action allAssetRequest instead of assetRequest
5) Replace assets will now use added logic:
   - If rendererUrl existence. If exists in adServer (preloaded) then check for renderAd availability on the window. If none exists, set load event on the script tag using the id 'pb-native-renderer'. If url sent as asset from Prebid, append script to window and attach load event. When script loaded and renderAd available, fire function to retrieve HTML. This inserted into innerHtml
   - Else if adTemplate exists, replace assets and insert into innerHtml
   - Else run existing replace method
6) replace function updated to include check for new native to replace keys which are in the format of ##KEY##

nativeAssetManager_spec and nativeTrackerManager_spec both updated to include additional tests

nativeTracker.js updated to include renderNativeAd function definition on window object


src/nativeTrackerManager.js
```diff
@@ -103,7 +122,30 @@ export function newNativeTrackerManager(win) {
...
+      console.warn('Prebid Native Tag object was missing \'adId\'.');
+    }
+
+    
+  }
+
   return {
-    startTrackers
+    startTrackers,
+    renderNativeAd
```
wqi1972, line 149: Suggest move `renderNativeAd` to a new class instead of add into `newNativeTrackerManager`. In long run, `newNativeTrackerManager.startTrackers` will get deprecated. If we separate code `renderNativeAd` from `newNativeTrackerManager` now, it will be much easier when that happens.

src/nativeTrackerManager.js
```diff
@@ -67,6 +68,24 @@ export function newNativeTrackerManager(win) {
...
 
+  function fireNativeImpTracker(adId){
+    let parsedUrl = parseUrl(window.pbNativeData.pubUrl);
+    publisherDomain = parsedUrl.protocol + '://' + parsedUrl.host;
+    fireTracker(adId, 'impression');
+  }
+
+  function fireNativeCallback(){
+    let parsedUrl = parseUrl(window.pbNativeData.pubUrl);
+    publisherDomain = parsedUrl.protocol + '://' + parsedUrl.host;
```
wqi1972, line 79: We suppose set `publisherDomain` value inside `renderNativeAd` once, instead of set it before each tracker call.

wqi1972:
I don't see code inside `renderNativeAd` take care of request for Mobile App as `startTrackers` does. I don't have experience on native ad for Mobile App, need someone has that experience double check it.

wqi1972:
We need logic to take care `````` and `````` as https://github.com/prebid/prebid-universal-creative/pull/60/files does.
This way will avoid 400 error message for icon and hero image.

src/nativeTrackerManager.js
mmoschovas: renderAd code separated into new nativeRenderManager

src/nativeTrackerManager.js
mmoschovas: Corrected the above to one definition in renderNativeAd as opposed to the tracker definitions

src/nativeAssetManager.js
```diff
@@ -160,14 +161,26 @@ export function newNativeAssetManager(win) {
...
+          if(key) {placeholders.push(key);}
+        });
+    }
 
     if (placeholders.length > 0) {
       callback = cb;
       requestAssets(adId, placeholders);
+    }else if(flag && win.pbNativeData.hasOwnProperty('requestAllAssets') && win.pbNativeData.requestAllAssets) {
+      callback = cb;
+      requestAllAssets(adId);
```
wqi1972, line 183: I think we suppose check `requestAllAssets` first. If it is true, then `requestAllAssets`. Only when `requestAllAssets` is not true, then `requestAssets` for  `placeholders`.

src/nativeTrackerManager.js
wqi1972: Good.

src/nativeAssetManager.js
mmoschovas: Can you describe the use case for this? I would think that if placeholders are detected/defined, then requesting all assets should not be a requirement as the needed assets would be known.

wqi1972: 2 reasons:
1) for Use Case 1 - “Render Replacements”, and Use Case 2 - “Ad Template as String”, these 2 use cases both have placementholder token inside html. pbNativeTagData.requestAllAssets = true; will be useless if the code check placementholders exist first;
2) for Use Case 1 - “Render Replacements”, for html `icon`, there suppose have some code which get asset of `##hb_native_icon##`, and set src of the image be the value of `##hb_native_icon##`. But inside html, there is no placeholder for `##hb_native_icon##`, so the asset won't get pulled in existing logic.

wqi1972: To make it easier, I will create a PR against your branch, so you will understand what I mean.

wqi1972: unfortunately my account can't push commit into git@github.com:mmoschovas/prebid-universal-creative.git repo, even inside the my own branch.
Here is the code change I made, I change function fireNativeCallback() to:
```
  function fireNativeCallback(){
    const adElements = findAdElements(AD_ANCHOR_CLASS_NAME);
    for (let i = 0; i < adElements.length; i++) {
      adElements[i].addEventListener('click', function(event) {
        loadClickTrackers(event, window.pbNativeData.adId);
      }, true);
    }
    const imgElements = win.document.getElementsByTagName("img");
    for (let i = 0; i < imgElements.length; i++) {
      if (imgElements[i].dataset && imgElements[i].dataset.src) {
        imgElements[i].src = imgElements[i].dataset.src;
      }
    }
  }
```
with this change, then it will support  tag with following format
```
<img data-src="##hb_native_image##" alt="" />
<img data-src="##hb_native_icon##" alt="logo" />
```
This will avoid 400 error if we directly set img.src inside html template.

wqi1972: Notice, this is a different solution than using ``, I think it is better and easier.
And with this solution, it doesn't need `pbNativeTagData.requestAllAssets = true;` for Use Case 1 - “Render Replacements” anymore, since all asset placementholder exists in GAM template html.

wqi1972: One more notice, you need to adjust `renderingMocks` inside nativeRenderManager_spec.js to 
```
const renderingMocks = {
  getWindowObject: function() {
    return {
      document: {
        getElementsByTagName: function() {
          return [];
        }
      },
      parent: {
        postMessage: sinon.spy()
      }
    }
  }
};
```

robertrmartinez:
Hey all, I am setting up some GAM line items today to help make testing local PUC changes easier.

I'll share example test pages once I am done today / tomorrow.

I would like to finalize the native stuff and merge both these for next weeks release!

Thanks again.

wqi1972:
I've tested on my local with following test cases:
1) New PBJS + PUC code with existing configure and templates, (using native key/value). Confirmed it is backward compatible; 
2) New PBJS + PUC code + sendTargetingKeys: false + template with macro ##native_<asset name>##. Confirmed  Use Case 1 - “Render Replacements” is working. The only issue the  tag temporarily raises 400 error. And with my suggestion provided above, that 400 error will get solved.

dist/load-cookie.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
```
robertrmartinez, line 1: I think this should not be checked in correct?

robertrmartinez: Same with the .DS_Store binary.

src/nativeAssetManager.js
```diff
@@ -227,12 +271,43 @@ export function newNativeAssetManager(win) {
...
     if (data.message === 'assetResponse') {
-      const body = win.document.body.innerHTML;
-      const newHtml = replace(body, data);
+      const body = win.document.body.innerHTML,flag = (typeof win.pbNativeData !== 'undefined');
 
-      win.document.body.innerHTML = newHtml;
-      callback && callback();
-      win.removeEventListener('message', replaceAssets);
+      if((data.hasOwnProperty('rendererUrl') && data.rendererUrl) || (flag && win.pbNativeData.hasOwnProperty('rendererUrl'))) {
+        if(win.renderAd){
```
robertrmartinez, line 277: some lint issues (we do not have automatic linter on puc code yet)

Maybe run the entire code on your local with your eslint plugin and do the fixes?

robertrmartinez, line 277: @jaiminpanchal27 When the rendererUrl is loaded, is the expected output that a renderAd function is placed on the win. 

If you could review this use case a little more in depth I would appreciate it since it was spec'd by you.

src/nativeRenderManager.js
```diff
@@ -0,0 +1,78 @@
...
+      if(nativeTag.hasOwnProperty('rendererUrl') && !nativeTag.rendererUrl.match(/##.*##/i)){
+        const scr = document.createElement('SCRIPT');
+        scr.src = nativeTag.rendererUrl,
+        scr.id = 'pb-native-renderer';
+        document.body.appendChild(scr);
+      }
+      nativeAssetManager.loadAssets(nativeTag.adId,fireNativeCallback);
+      fireNativeCallback();
+      fireNativeImpTracker(nativeTag.adId);
+    }else{
```
robertrmartinez, line 70: some more lint issues

src/nativeTrackerManager.js
```diff
@@ -10,6 +10,7 @@ const AD_DATA_ADID_ATTRIBUTE = 'pbAdId';
 export function newNativeTrackerManager(win) {
   let publisherDomain;
 
+
```
robertrmartinez, line 13: extra new line?

test/spec/nativeTrackerManager_spec.js
```diff
@@ -89,6 +95,58 @@ describe('nativeTrackerManager', function () {
       expect(trimPort(postMessageTargetDomain)).to.equal(tagData.pubUrl);
     });
 
+    /*it('should verify the postMessage for impression trackers was executed', function() {
```
robertrmartinez, line 98: what is the reasoning behind this being commented out?

src/nativeTrackerManager.js
mmoschovas: Removing this. I believe I originally added logic that I later removed.

test/spec/nativeTrackerManager_spec.js
mmoschovas: I forgot to remove this test. Originally the new rendering code lived within trackerManager code but it was requested to live in a new class. This was mirrored in the nativeRenderManager_spec test.

src/nativeAssetManager.js
mmoschovas: Updated

src/nativeRenderManager.js
mmoschovas: Updated

robertrmartinez:
Ok changes look good to me.

Would like @jaiminpanchal27 to do a review before we merge this.

Specifically making sure the Xander proposal of rendererUrl looks good. since they spec'd that part out.

More than happy to set some time or keep an open dialogue if you want @jaiminpanchal27 

Thanks all!

robertrmartinez:
The associated Prebid.js change has been approved and is ready for merge once this has been approved and merged as well:

https://github.com/prebid/Prebid.js/pull/5411

robertrmartinez (approved):
Looks good to me

jsnellbaker (approved):
@mmoschovas 
I think the code here looks okay.  Can you please remove the `.DS_Store` file and the `package-lock.json` updates from the PR though?

mmoschovas:
Removed both updates

wqi1972 (approved):

src/nativeAssetManager.js
```diff
@@ -227,12 +271,45 @@ export function newNativeAssetManager(win) {
     }
 
     if (data.message === 'assetResponse') {
-      const body = win.document.body.innerHTML;
-      const newHtml = replace(body, data);
+      const body = win.document.body.innerHTML,flag = (typeof win.pbNativeData !== 'undefined');
```
robertrmartinez, line 274: Can you split this `flag` out to a new declaration please.

src/nativeAssetManager.js
```diff
@@ -227,12 +271,45 @@ export function newNativeAssetManager(win) {
...
 
-      win.document.body.innerHTML = newHtml;
-      callback && callback();
-      win.removeEventListener('message', replaceAssets);
+      if (flag && data.adId != win.pbNativeData.adId) return;
```
robertrmartinez, line 276: Any particular reason to use `!=` vs `!===`?

Do we expect adId to be the same data type in both? Or is is possible one is string and one is number?

src/nativeRenderManager.js
robertrmartinez: Should have a space after closing bracket befor else and after

```javascript
} else {
```

src/nativeAssetManager.js
```diff
@@ -190,6 +204,19 @@ export function newNativeAssetManager(win) {
...
+   */
+  function scanForString(str) {
+    const strIndex = win.document.body.innerHTML.indexOf(str);
+
+    if (~strIndex) {
+      return true;
+    }
+
+    return false;
+  }
```
robertrmartinez, line 218: Could just do

```javascript
return win.document.body.innerHTML.indexOf(str) !== -1
```


Not sure this function is even being used?

src/nativeAssetManager.js
```diff
@@ -160,14 +161,26 @@ export function newNativeAssetManager(win) {
...
+        win.pbNativeData.assetsToReplace.forEach((asset) => {
+          const key = (asset.match(/hb_native_/i)) ? asset : NATIVE_KEYS[asset];
+          if (key) {placeholders.push(key);}
+        });
+    }
 
     if (placeholders.length > 0) {
       callback = cb;
       requestAssets(adId, placeholders);
+    }else if (flag && win.pbNativeData.hasOwnProperty('requestAllAssets') && win.pbNativeData.requestAllAssets) {
```
robertrmartinez, line 181: space after closing brackets

src/nativeRenderManager.js
robertrmartinez: maybe just run enable eslint in your editor and see what else there is?

robertrmartinez (changes requested):
couple minor things

src/nativeAssetManager.js
mmoschovas: Done

src/nativeAssetManager.js
mmoschovas: Nope good catch. Updated

src/nativeAssetManager.js
mmoschovas: I dont remember this function but cant find a use for it so removing

src/nativeAssetManager.js
mmoschovas: Fixed these. Not sure how they didnt update last time around

bretg:
Docs PR https://github.com/prebid/prebid.github.io/pull/2479

allanjun (approved):

jsnellbaker:
@mmoschovas Can you please take a look at the code conflict in the `gulpfile.js`?  I'm not sure if it's a recent thing, but it needs to be addressed.

Thanks.

mmoschovas:
@jsnellbaker committed merge

jsnellbaker:
@mmoschovas Thanks for the updates and for the overall effort behind this PR.  Based on side conversations, we are good to merge this PR (finally).

PR #111, opened 2020-07-24
Fix amp cache size issue

idettman:


src/renderingManager.js
```diff
@@ -156,7 +158,14 @@ export function newRenderingManager(win, environment) {
     // For MoPub, creative is stored in localStorage via SDK.
     let search = 'Prebid_';
     if(uuid.substr(0, search.length) === search) {
-      loadFromLocalCache(uuid)
+      loadFromLocalCache(uuid);
+      //register creative right away to not miss initial geom-update
+      if (typeof size !== 'undefined' && size !== "") {
```
harpere, line 163: wouldn't `if (size) {` work just as well here?

src/renderingManager.js
```diff
@@ -156,7 +158,14 @@ export function newRenderingManager(win, environment) {
...
+        let sizeArr = size.split('x').map(Number);
+        resizeIframe(sizeArr[0], sizeArr[1]);
+      } else {
+        console.log('Targeting key hb_size not found to resize creative');
+      }
     } else {
```
harpere, line 169: this resize code is a duplicate of what's above.  can we consolidate?

jsnellbaker (approved):

PR #115, opened 2020-08-20
cookie_sync update for bidder priority

mmoschovas:
Update to retrieve bidder list from url query params and pass them along as new data property to sync endpoint


Fawke (approved):
LGTM

PR #124, opened 2020-12-02
update wdio dev deps

ReinoutStevens:
This fixes https://github.com/prebid/prebid-universal-creative/issues/120

Tests still run locally.
I guess circleci will tell me if browserstack still works or not (edit: there is no circleci here, so no idea )


Fawke (commented):
@ReinoutStevens,

Thanks for the PR. This update was due for a long time, but, the command `gulp e2e-test` fails because there's an API change in webdriver@6.

I created another PR because we needed to make some changes in other places as well. https://github.com/prebid/prebid-universal-creative/pull/132

jsnellbaker:
Closing this as a duplicate to other tickets (possibly #132 but also #159 )

PR #142, opened 2021-05-13
Add MRAID viewability support to bURL trigger

kizzard:
This adds MRAID 1-3 viewability support for triggering the bURL. This is an attempt to cut down discrepancies. Note this is my first time writing MRAID code so would appreciate a thorough review if anyone has expertise.

### DIFF
--- a/src/renderingManager.js
+++ b/src/renderingManager.js
@@ -213,6 +213,22 @@ export function newRenderingManager(win, environment) {
         if (bidObject.nurl) {
           ad += utils.createTrackPixelHtml(decodeURIComponent(bidObject.nurl));
         }
+        if (bidObject.burl) {
+          let triggerBurl = function(){ utils.triggerPixel(bidObject.burl); };
+          if(isMobileApp) {
+            let mraidScript = utils.loadScript(win, 'mraid.js',
+              function() { // Success loading MRAID
+                let result = registerMRAIDViewableEvent(triggerBurl);
+                if (!result) {
+                  triggerBurl();
+                }
+              },
+              triggerBurl // Error loading MRAID
+              );
+          } else {
+            triggerBurl();
+          }
+        }
         utils.writeAdHtml(ad);
       } else if (bidObject.nurl) {
         if(isMobileApp) {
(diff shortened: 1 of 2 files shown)


idettman (approved):
LGTM

kizzard:
@idettman thanks for the review. Any timeline for a possible merge?

mmullin:
@idettman Can we get a merge here?

src/renderingManager.js
```diff
@@ -213,6 +213,22 @@ export function newRenderingManager(win, environment) {
...
+          let triggerBurl = function(){ utils.triggerPixel(bidObject.burl); };
+          if(isMobileApp) {
+            let mraidScript = utils.loadScript(win, 'mraid.js',
+              function() { // Success loading MRAID
+                let result = registerMRAIDViewableEvent(triggerBurl);
+                if (!result) {
+                  triggerBurl(); // Error registering event
+                }
+              },
+              triggerBurl // Error loading MRAID
```
mkendall07, line 226: I don't think `loadScript` takes a 4th param for error handling.

kizzard:
thanks @mkendall07 . Can we get a release? We developed this patch as a fix for a publisher using the CDN hosted release.

mkendall07:
Oh Hey Jeremy - didn't realize that was you :). @jsnellbaker usually does these releases so just tagging him here.

kizzard:
Thanks Matt (& Jason)!

bretg:
@kizzard - where is mraid.js supposed to come from? We suspect this is causing problems because it's not available in all environments.

kizzard:
@bretg As per spec, I believe the rendering SDK or delivery ad server should provide mraid.js. From the 3.0 spec:

> 2.2 Ad Control 
Ad designers that expect ads to make use of MRAID must invoke the mraid.js 
script before the ad is loaded (see section 3 on initialization). Once called, the ad 
container can inject the MRAID JavaScript into the ad file.

The creative is not expected to provide it. The code in this MR is designed to fall back gracefully if mraid.js is not available or otherwise fails to load, and uses a simple pixel trigger instead of MRAID. There are several layers of fallback functionality provided.

Back in the day, the ADTECH ad server used to literally scan the creative code for the string `mraid.js` to decide whether or not to inject it - so perhaps burying it in a CDN hosted file is not helping matters, but I would have expected Google's detection to be more advanced.

As an aside, I have debugged a HAR log of a failing MRAID creative delivered by PBS on IOS provided by a client (probably a similar issue you are hearing about) and running it locally, the MRAID error was being triggered in the creative's mraid code itself, not the PUC burl code which fell back to basic pixel triggering. Just an anecdote but thought I would mention it.

PR #148, opened 2021-11-29
Static vast tag for mobile rendering API for rewarded video

YuriyVelichkoPI:
## Motivation

Host a static VAST tag in order to use it instead of PUC in the scenario for integration of the Prebid Mobile Rendering API with GAM Rewarded Video ads.

## Description of the tag

This tag signals the Prebid Mobile SDK that Prebid's Line Item for the Rewarded Video wins. The trigger is a tag

```
<AdTitle>PrebidAppEvent</AdTitle>
```

Based on it the SDK determines a winner and manages to render it.

## Future Steps

Need to host this file in the public space in order to use its URL in the GAM Creative setup.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -155,6 +155,11 @@ function buildUid() {
     .pipe(gulp.dest('dist'));
 }
 
+function includeStaticVastXmlFile() {
+  let target = gulp.src('static/prebid-mobile-rewarded-vast.xml');
+  return target.pipe(gulp.dest('dist'));
+}
+
 // Run the unit tests.
 //
 // By default, this runs in headless chrome.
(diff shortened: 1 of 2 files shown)


bretg:
The goal is to host this on as https://cdn.jsdelivr.net/npm/prebid-universal-creative@latest/dist/prebid-mobile-rewarded-vast.xml

@jsnellbaker and @ChrisHuie - please coordinate for how to make that happen. I believe there's a Jenknins job associated with this repo like there is for PBJS. Thanks.

jsnellbaker (changes requested):
Hi,

if we include moving this file into the `gulp build` process, subsequent releases of PUC will ensure the latest version of the file would be hosted at the location @bretg outlined on the jsdevlier domain.

From my quick test, I believe some new function like the following would need to be added to the `gulpfile.js` and then call that function as part of the main `build` gulp task found later down in the file (it can be added to the comma separated list with all the `buildXYZ` commands).

```
function includeStaticVastXmlFile() {
  let target = gulp.src('static/prebid-mobile-rewarded-vast.xml');
  return target.pipe(gulp.dest('dist'));
}
```

```
gulp.task('build', gulp.parallel(buildProd, buildCookieSync, buildCookieSyncWithConsent, buildNative, buildNativeRender, buildUid, includeStaticVastXmlFile));
```

If you then run `gulp build` locally on the project, you should see a copy of the file moved into the `dist` that's created/populated with all the normal build files.  

If this updated `gulp build` command works locally, it should work the same for the Jenkins job.

jsnellbaker (approved):
I have pushed a commit on the fork to handle the proposed gulpfile updates.

Need a 2nd review.

harpere (approved):

PR #154, opened 2022-01-21
Use MessageChannel for cross-frame messages

dgirardi:
Use MessageChannels to enable communication from behind an opaque origin

See https://github.com/prebid/Prebid.js/issues/7870
Related PBJS changes: https://github.com/prebid/Prebid.js/pull/7953

### DIFF
--- a/src/messaging.js
+++ b/src/messaging.js
@@ -1,0 +1,43 @@
+import {parseUrl} from './utils.js';
+
+export function prebidMessenger(publisherURL, win = window) {
+    const prebidDomain = (() => {
+        if (publisherURL == null) {
+            return null;
+        }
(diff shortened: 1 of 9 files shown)


dgirardi:
Tested native integration using this test page: https://gist.github.com/dgirardi/7525bc3d9f7b92f3ff919fc10ede81d6 (expects PUC to live at localhost:9990)

That covers the render-native.js case; the only example I could find for native-trk.js is meant for mobile (and wouldn't work because no `pubUrl` is defined in the creative)

patmmccann:
@msm0504 are you able to take a look?

robertrmartinez (approved):
Looks good!

PR #159, opened 2022-02-23
Upgrade dependencies make the build work on recent(ish) versions of Node

dgirardi:
Upgrade dependencies to make the build work on node v14 (used by https://github.com/prebid/prebidjs-releaser﻿)

With these changes, `npm ci` and gulp `build`, `test` and `serve` still work on node v8.


jsnellbaker (approved):

jsnellbaker:
@dgirardi Thanks for putting this together; not sure if we need a 2nd review in this case.

PR #165, opened 2022-03-15
Separate puc

musikele:
With this PR when puc is built (`gulp build`) output of that task will be multiple `.js` file. Each `.js` file is responsible for rendering ad of specific format:
- amp.js
- banner.js
- mobile.js
- native.js
- video.js

Before, when puc was built, in the output we would have:
- creative.js (for all formats except native)
- native-render.js (only for native format)

Not a single function was changed, all functions were just moved around (some to different files).  
Also, some OOP concepts were removed to make treeshaking easier.


musikele:
Docs PR: https://github.com/prebid/prebid.github.io/pull/3645

.gitignore
```diff
@@ -1,4 +1,5 @@
 # Built files
 node_modules/
 build
-coverage
\ No newline at end of file
+coverage
+dist/
```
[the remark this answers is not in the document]
musikele: added file .npmignore

gulpfile.js
```diff
@@ -106,19 +153,88 @@ function buildProd() {
...
     }))
     .pipe(gulp.dest('dist'));
 }
 
+function includeStaticVastXmlFile() {
+  let target = gulp.src('static/prebid-mobile-rewarded-vast.xml');
+  return target.pipe(gulp.dest('dist'));
+}
+
+function buildBanner() {
```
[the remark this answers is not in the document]
musikele: created two generic functions, buildDev and buildProd, and other functions will call these.

dgirardi (approved):
LGTM with the limited testing I was able to do: both `dist/creative.js` and `dist/banner.js` work as expected with simple banner ads. `dist/video.js` is an exact copy of `banner.js`. I am completely ignorant of the mobile / AMP flow.

jsnellbaker:
@musikele When you get a chance, can you review/resolve the conflicts in this PR?  Thanks.

gulpfile.js
```diff
@@ -151,7 +190,7 @@ function buildUid() {
   return gulp.src(['src/ssp-userids/uid.js'])
   .pipe(webpackStream(cloned))
     .pipe(uglify())
-    .pipe(header('/* v<%= creative.version %>\n' + dateString + ' */\n', { creative: creative }))
+    .pipe(header('/* v<%= creative.version %>\n' + dateString + '\nDEPRECATED, please use creative based on hb_format targeting */\n', { creative: creative }))
```
jsnellbaker, line 193: Is this comment in the right prod build function?  I might have missed it, but what hb_format does the uid.js file connect with?  Was this supposed to go in one of the older native prod files?

src/nativeRenderManager.js
```diff
@@ -48,25 +48,25 @@ export function newNativeRenderManager(win) {
   }
 
   // START OF MAIN CODE
-  let renderNativeAd = function(nativeTag) {
+  let renderNativeAd = function(doc, nativeTag) {
     window.pbNativeData = nativeTag;
     sendMessage = prebidMessenger(nativeTag.pubUrl, win);
     const nativeAssetManager = newNativeAssetManager(window, nativeTag.pubUrl);
 
     if (nativeTag.hasOwnProperty('adId')) {
```
jsnellbaker, line 56: When I was running a basic test where I had a GAM creative setup with the new basic template (from the updated README.md) and calling a native creative, the new logic for the `native.js` ran into an issue here b/c the `adId` is not part of the `nativeTag` object that gets passed into the creative's renderAd() function; the adId is still stuck inside the targetingMap param in its natural format.  See below for a copy of the creative code in its initial state when it gets to the code at this point:
```
  var ucTagData = {};
  ucTagData.adServerDomain = "";
  ucTagData.pubUrl = "http://test.localhost:9999/integrationExamples/gpt/hello_world_puc_adformat_nonsf.html?pbjs_debug=true";
  ucTagData.targetingMap = {"hb_adid":["3b4e49b4f268f3"],"hb_adid_appnexus":["3b4e49b4f268f3"],"hb_bidder":["appnexus"],"hb_bidder_appnexus":["appnexus"],"hb_format":["native"],"hb_format_appnexus":["native"],"hb_pb":["5.00"],"hb_pb_appnexus":["5.00"],"hb_size":["0x0"],"hb_size_appnexus":["0x0"]};
  ucTagData.requestAllAssets = true;

  try {
    ucTag.renderAd(document, ucTagData);
  } catch (e) {
    console.log(e);
  }
```  

Because it can't find the `adId`, the rest of the rendering process is aborted.

In the previous way with the `native-render.js` approach, the `adId` a direct part of the object that's passed to the render function, see below for reference:
```
let pbNativeTagData = {};
pbNativeTagData.pubUrl = "https://sneezy.aws.rubiconproject.com/rp-eng-soln/prebid/prebid_1001_native_case4.html?pbjs_debug=true";
pbNativeTagData.adId = "3b8572ff0e96ab";
pbNativeTagData.requestAllAssets = true;
window.pbNativeTag.renderNativeAd(pbNativeTagData);
```

README.md
```diff
@@ -30,6 +30,16 @@ You can find a detailed explanations on the [Prebid Universal Creative](http://p
 </script>
 ```
 
+Creative created like described above will work for all formats:
+- amp
+- banner
+- mobile
+- native
```
jsnellbaker, line 37: From some initial tests I was running with a new GAM creative using the above code template, I'm not sure if simply writing 'native' here is accurate.  From what I can tell, it seems like not all native setups that we support will be handled with just the code above. 

Specifically the scenario where the native template is set within the creative's adserver.  In that case, the shell of the creative's HTML is included above the script tag that loads the PUC native script (whether it's the new `native.js` or the previous incarnations of `native-render.js` or `native-trk.js`).
https://docs.prebid.org/prebid/native-implementation.html#native-template-in-adserver

If that's indeed the case, then the text here should highlight the restrictions (or give a link to the prebid.org native docs that can better explain what scenarios have specific requirements).

src/nativeRenderManager.js
```diff
@@ -48,25 +48,25 @@ export function newNativeRenderManager(win) {
   }
 
   // START OF MAIN CODE
-  let renderNativeAd = function(nativeTag) {
+  let renderNativeAd = function(doc, nativeTag) {
```
jsnellbaker, line 51: While running a regression test on the using the `native-render.js` type creative setup, I encountered an undefined error when the creative started to call the `pbNativeTag.renderNativeAd()` via the new `legacyNativeRender.js` file.

The issue is in the call signature for this function here.  Previously there was a single argument for the `nativeTag` which contained the pubUrl and other info.  Now with the updated signature, this data getting passed into the `doc` property instead of the `nativeTag` 2nd param, which breaks the rest of the logic in this function.

We need some new logic here to handle the old scenario since this file is shared between the old/new native rendering routes.

robertrmartinez:
Hi when I pull down this and try `npm ci` I get the following error:

image

Prob because I am using NODE 12, when I use node 16 it works.

So I guess not a big deal but maybe we should commit the changes to `package-lock.json` as well?

src/nativeRenderManager.js
musikele: Hi @jsnellbaker , the error is in the readme.md file. The right creative should be: 

```
<script src = "https://cdn.jsdelivr.net/npm/prebid-universal-creative@latest/dist/%%PATTERN:hb_format%%.js"></script>
<script>
  var ucTagData = {};
  ucTagData.adServerDomain = "";
  ucTagData.pubUrl = "%%PATTERN:url%%";
  ucTagData.targetingMap = %%PATTERN:TARGETINGMAP%%;
  ucTagData.hbPb = "%%PATTERN:hb_pb%%";
  ucTagData.hbFormat = "%%PATTERN:hb_format%%";
  ucTagData.adId = "%%PATTERN:hb_adid%%";
  ucTagData.requestAllAssets = true;

  try {
    ucTag.renderAd(document, ucTagData);
  } catch (e) {
    console.log(e);
  }
</script>
```

Going to update the readme file.

README.md
musikele: I've fixed the [documentation PR](https://github.com/prebid/prebid.github.io/pull/3645/commits/24b5e38ba896d3fda37e2acd632dad57e6f48538) to specify the `document` object as a first argument of the `renderAd` function. 

In general, I don't really like the first scenario of the native approach, where the template is specified in GAM; it looks to me as something that's being carried on for compatibility with the past. It's also the scenario that is worse to customize because it requires to access GAM at every change. Anyway, the PR is compatible with that scenario too; i just tested it. Thanks for catching the mismatch.

src/nativeRenderManager.js
musikele: I changed file `legacyNativeRender` so the `renderNativeAd` function that is exposed accepts only one parameter. I also added two tests for it.

musikele:
@robertrmartinez Added package-lock.json to fix `npm ci`.

src/nativeRenderManager.js
jsnellbaker: Hi @musikele 

When I tried to test this out again on a test page, I was still seeing the behavior.  I attached a screenshot for reference.
I saw it attempt to use the new call function you added, but I think the call interpreted the `document` variable as the `this` context for the call function rather an argument to pass through to the subsequent function.  Can you take a look again?


Screen Shot 2022-09-27 at 10 08 02 AM

jsnellbaker: Confirmed fix per the latest updates.

jsnellbaker (approved):
@musikele 

Thank you for keeping this PR healthy and making the tweaks to the things I found.  TMK with what tests I was able to run, it seems like old regression files and the new files appear to be working.  With proper documentation for the different usages, this should hopefully help people transition.

patmmccann:
@musikele looks like we're ready to merge once conflicts are resolved?

GLStephen:
@patmmccann came by to check up on things after our discussion today, is the README referencing https://cdn.jsdelivr.net/npm/prebid-universal-creative@latest/dist/%%PATTERN:hb_format%%.js out of sync with actual jsdeliver deployment right now? https://cdn.jsdelivr.net/npm/prebid-universal-creative@latest/dist/banner.js is not available. Looks like there is just some flux in state of things, but wanted to bring it up if it mattered.

PR #166, opened 2022-04-15
Remove doctype comments from markup. 

patmmccann:
Attempted solution to https://github.com/prebid/prebid-universal-creative/issues/134


patmmccann:
see https://jsfiddle.net/ayxos/k7p14c3e/3/

PR #171, opened 2022-08-10
Replace missing native assets with empty strings

anitaschiller:
See the following [issue](https://github.com/prebid/Prebid.js/issues/8769) and closed [PR](https://github.com/prebid/Prebid.js/pull/8773) for Prebid.js. 

Instead of replacing missing native assets in Prebid Core, all remaining placeholders in the template are now directly replaced by empty strings in the `nativeAssetManager`. 

After replacing the placeholders with existing assets, all placeholders available in the original template are searched for again and if still found, replaced by an empty string. 

I also extracted the `typeof win.pbNativeData !== 'undefined'` flag to a function. I'm not 100% sure what it does but I suppose it checks if the template is a GAM template? I hope I used it the right way.

Thanks for your review!

### DIFF
--- a/src/nativeAssetManager.js
+++ b/src/nativeAssetManager.js
@@ -70,8 +70,6 @@ export function newNativeAssetManager(win, pubUrl) {
     }
   }
 
-
-
   function getCacheEndpoint(cacheHost, cachePath) {
     let host = (typeof cacheHost === 'undefined' || cacheHost === "") ? DEFAULT_CACHE_HOST : cacheHost;
     let path = (typeof cachePath === 'undefined' || cachePath === "") ? DEFAULT_CACHE_PATH : cachePath;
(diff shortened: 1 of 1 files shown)


src/nativeAssetManager.js
```diff
@@ -333,17 +336,24 @@ export function newNativeAssetManager(win, pubUrl) {
...
     (assets || []).forEach(asset => {
-      const flag = (typeof win.pbNativeData !== 'undefined');
+      const flag = pbNativeDataHasValidType();
       const searchString = (adId && !flag) ? `${NATIVE_KEYS[asset.key]}:${adId}` : ((flag) ? '##'+`${NATIVE_KEYS[asset.key]}`+'##' : `${NATIVE_KEYS[asset.key]}`);
       const searchStringRegex = new RegExp(searchString, 'g');
       html = html.replace(searchStringRegex, asset.value);
     });
 
+    scanForPlaceholders().forEach(placeholder => {
+      const searchString = pbNativeDataHasValidType() ? `##${placeholder}##` : placeholder;
```
dgirardi, line 353: I'm not sure this is correct, because the logic on how to construct the placeholder is different from what's just a few lines above in `replace`. (Unfortunately I'm not sure what the logic _should_ be either, it looks like it supports yet more obscure use cases).

Would it be better to move this inside `replace`? on line 345 you could loop over `scanForPlaceHolders()` instead of `assets`.

anitaschiller: @dgirardi I get your point!

I flipped it like you suggested, now looping over `scanForPlaceholders` with more or less the original logic. Is this what you were thinking of?

src/nativeAssetManager.js
```diff
@@ -333,16 +334,18 @@ export function newNativeAssetManager(win, pubUrl) {
...
+   * If there's no actual value, the placeholder gets replaced by an empty string.
    */
   function replace(document, { assets, adId }) {
     let html = document;
 
-    (assets || []).forEach(asset => {
-      const flag = (typeof win.pbNativeData !== 'undefined');
-      const searchString = (adId && !flag) ? `${NATIVE_KEYS[asset.key]}:${adId}` : ((flag) ? '##'+`${NATIVE_KEYS[asset.key]}`+'##' : `${NATIVE_KEYS[asset.key]}`);
+    scanForPlaceholders().forEach(placeholder => {
+      const flag = true;
```
dgirardi, line 343: yes, except why is this fixed to true now?

anitaschiller: Oh, that's a mistake, sorry! I fixed it.

dgirardi (approved):

patmmccann (approved):

PR #180, opened 2022-09-30
Integration tests

dgirardi:
Integration tests; this aims to be exhaustive, but is a WIP.

Cases covered so far:

- Banner (should display, fire won and render events, substitute macros) 
- Banner on AMP (should display, substitute macros - see #183) , trigger nurl/burl/wurl)
- Native using legacy adUnit definitions and templates (should display, substitute asset macros, fire won events, fire trackers. Also see #182 )
- All of the above is repeated in and out of safeframes

In progress:
- New style (ortb) native adUnits and templates
- Banner on app/mobile 

Missing:
- `native-trk.js` (need examples/docs!)
- mediaType-specific PUC builds (waiting for https://github.com/prebid/prebid-universal-creative/pull/165)

Not supported by PUC (!) 

- Native on AMP or app


musikele:
The biggest issue with this PR is that running tests is aleatory. Basically, all tests may fail because of network, cpu, bandwith, google deciding that we're overcalling their endpoint, and other thousand million causes that it'll be impossible to say "oh, tests are failing because it's my fault". 

Another issue is that these tests take several minutes to run. On a local pc they may take up to 5-6 minutes. So they're only suitable for a CI environment. 

So, probably if we remove some tests, or some browsers, and we keep the run timing under 2 minutes, and we tweak the parameters so that tests failing go to zero, this is the test suite we were missing

PR #188, opened 2022-11-07
make the project work with node latest; upgrade wdio

musikele:
prebid-universal-creative had, as a requirement, node <= 8.7.0 (Released: 2017-10-11). 

With this PR I have updated WDIO to latest version and relaxed the constraint about node version (now it's generically >= 8.7.0).


dgirardi:
LGTM, but also, I'm removing wdio entirely in https://github.com/prebid/prebid-universal-creative/pull/180/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519

musikele:
So we are not merging this in favour of the playwright-test merge? 

What if I update this PR to remove wdio entirely too? I guess browserstack won't run anymore?

PR #189, opened 2022-11-07
add build-dev task

musikele:
Currently, if a user wants to generate dev files (that is, unminified files together with source maps), he should either call `npx gulp test` or `npx gulp serve`. Both tasks generate dev files but they also perform other tasks that may be irrelevant for those who only want to generate dev files. 

With this PR, I add a new task `npx gulp build-dev` that will only generate dev files.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -244,7 +244,9 @@ function setupE2E(done) {
 
 gulp.task('test', gulp.series(clean, test));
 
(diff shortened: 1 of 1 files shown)


jsnellbaker (approved):
LGTM

Not sure if we really need a 2nd review here, so will leave it un-merged for the time being.

musikele:
@dgirardi can you give a look at this PR too?

dgirardi (approved):

PR #201, opened 2023-01-23
Native click trackers: track all clicks (including middle and right button clicks)

dgirardi:
Closes https://github.com/prebid/prebid-universal-creative/issues/198


patmmccann (approved):

patmmccann (changes requested):
just noting touch events may not fire, subject to testing

patmmccann (approved):

PR #211, opened 2023-04-26
Replace all placeholders with the corresponding ortb assets

anitaschiller:
At the moment **only the first occurrence** of an ortb placeholder **is replaced by the asset value**, the second one will be replaced by an empty string. 

I found this issue because our native design template uses the same placeholder twice for responsiveness reasons. Depending on the viewport size, one of them has the CSS property `display: none` but they are both present in the template. When the second occurrence gets displayed, the html element is always empty - I would expect all placeholders with the corresponding asset id to be replaced though.

Thank you for checking! 😊

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "description": "Universal creative for Prebid apps",
     "main": "dist/creative.js",
     "scripts": {
-        "test": "echo test",
+        "test": "gulp test",
         "prepublish": "gulp build"
     },
     "repository": {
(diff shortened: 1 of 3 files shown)


patmmccann (approved):

musikele (changes requested):
Hi there. I request only one change: to add a test case. 

It'll be very easy, infact I have just wrote it for you :D 

in nativeAssetManager_spec.js , in test "no placeholders found but requests all assets flag set - adTemplate - openRTB", add a new placeholder (for example I added `<h2>##hb_native_asset_id_1##</h2>`, that corresponds to `new value` ) and then test with 

```
// test that we are replacing ALL assets occurrences occurrences 
expect(([...win.document.body.innerHTML.match(/new value/g)] || []).length, "expected 2 occurrences of \"new value\"").to.equal(2);
```

musikele (approved):
LGTM

PR #217, opened 2023-08-19
Log error when renderAd lookup failed

Sir-Will:
When the universal creative fails to find the `renderAd` function, it silently continues, resulting in a failed ad render not being noticeable in the console. This adds an error message when it happens.


patmmccann (approved):

Sir-Will:
Thanks, @patmmccann.
After a second review, I noticed it doesn't fully do its job. https://github.com/prebid/prebid-universal-creative/pull/218

PR #233, opened 2024-06-21
Adding some comments to nativeAssetManager.js

bretg:
Thought these might help the next poor soul who has to try to figure this thing out

### DIFF
--- a/src/nativeAssetManager.js
+++ b/src/nativeAssetManager.js
@@ -333,11 +333,17 @@ export function newNativeAssetManager(win, nativeTag, mkMessenger = prebidMessen
           renderPayload.ortb = data.ortb;
         }
 
+        // if there's a rendererUrl, we need to check whether it's already been loaded.
+        // The mark of it having been loaded is the existence of a window.renderAd() function.
+        // There are 3 scenarios:
+        //   1) it's already been loaded in some undocumented way by a publisher or the creative
+        //   2) it was already loaded from nativeRenderManager.js (which creates a script with is pb-native-renderer)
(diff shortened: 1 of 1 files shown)


src/nativeAssetManager.js
```diff
@@ -333,11 +333,17 @@ export function newNativeAssetManager(win, nativeTag, mkMessenger = prebidMessen
           renderPayload.ortb = data.ortb;
         }
 
+        // if there's a rendererUrl, we need to check whether it's already been loaded.
+        // The mark of it having been loaded is the existence of a window.renderAd() function.
+        // There are 3 scenarios:
+        //   1) it's already been loaded in some undocumented way by a publisher or the creative
+        //   2) it was already loaded from nativeRenderManager.js (which creates a script with is pb-native-renderer)
+        //   3) it hasn't been loaded yet
```
dgirardi, line 341:
```suggestion
        // if there's a rendererUrl, we need to check whether it's already been loaded.
        // There are 3 scenarios:
        //   1) it's already been loaded (window.renderAd is present)
        //   2) it is currently being loaded (through a script tag with id "pb-native-renderer")
        //   3) it hasn't been loaded yet
        //  1 and 2 seem intended to work with logic in nativeRenderManager.js, which (sometimes) loads rendererUrl through a <script id="pb-native-renderer">, but it could conceivably be used in an undocumented way to embed renderer logic directly in the creative.
```

dgirardi (approved):

muuki88:
It will indeed :tada:

patmmccann (approved):

PR #244, opened 2024-11-06
Add Open Measurement impression tracking

ValentinPostindustria:
We have researched the Open Measurement SDK's behavior for impression tracking (`burl`) in the PUC. 

OM SDK JS script contains the public methods that are available in all versions and are not obfuscated. We have researched different versions of AppLovin, GAM, and AdMob. List of these methods:
```
startSession
error
finishSession
publishAdEvent
publishImpressionEvent
publishVideoEvent
publishMediaEvent
publishLoadedEvent
setNativeViewHierarchy
setState
setDeviceVolume
init
setLastActivity
```
We can subscribe to all these methods using the proxy method. For receiving them we should use `window.omidBridge.<method_name>`. 

We have tried to subscribe to `publishImpressionEvent`, but it works too fast and we can't handle it. So we decided to subscribe to the geometry change called by the `setNativeViewHierarchy` method. But this method can be called when a view is not on a screen or is not visible, so we need to check this method's parameter. Parameter options:

- [gc_view_is_not_on_screen.json](https://github.com/user-attachments/files/17660528/gc_view_is_not_on_screen.json) - checked by PR method `viewIsNotOnScreen`.
- [gc_view_is_not_visible.json](https://github.com/user-attachments/files/17660526/gc_view_is_not_visible.json) - checked by PR method `viewIsNotVisible`.
- [gc_view_visible.json](https://github.com/user-attachments/files/17660524/gc_view_visible.json) - case when we can track impression.

### DIFF
--- a/src/mobileAndAmpRender.js
+++ b/src/mobileAndAmpRender.js
@@ -6,6 +6,9 @@ import { writeAdHtml } from './postscribeRender';
 const DEFAULT_CACHE_HOST = 'prebid.adnxs.com';
 const DEFAULT_CACHE_PATH = '/pbc/v1/cache';
 
+let omImpressionTracked = false;
+let burlTracked = false;
+
 /**
  * Render mobile or amp ad
  * @param {string} cacheHost Cache host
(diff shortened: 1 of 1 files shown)


bretg:
Thanks @ValentinPostindustria and @YuriyVelichkoPI  -- while you're working on PUC tracking, please take a look at https://github.com/prebid/prebid-mobile-ios/issues/1059

jclou:
Some concerns based on my understanding of Open Measurement:
1. If the ad doesn't contain any verifications or if GAM/AdMob etc. choose not to run any verifications (e.g. the verification provider is not allowlisted), then in an optimized integration there is no reason to set up an OMSDK ad session and invoke the methods we are intercepting.
2. The omid bridge in the service script is an implementation detail, designed to be be a private API between OMSDK native and javascript layers -- and therefore is subject to change without backwards compatibility.

Would it make more sense to instead write a proper verification script and use omid as intended?

rickyblaha:
Please add some automated tests to cover these changes. For example see some added here for rendering based on MRAID viewability https://github.com/prebid/prebid-universal-creative/pull/248

YuriyVelichkoPI:
Closing this PR in favor of more reliable approaches: 
- [Native Viewability Tracking](https://github.com/prebid/prebid-mobile-ios/pull/1060) 
-[ Custom OM script](https://github.com/prebid/prebid-mobile-ios/issues/1095)

PR #245, opened 2024-11-09
Fix mobile-app rendering when admarkup contains lowercase DOCTYPE tag

rickyblaha:
Admarkup containing an html DOCTYPE tag `<!DOCTYPE html>` prevents rendering with postscribe.

The uppercase tag did get removed since https://github.com/prebid/prebid-universal-creative/pull/210, but the lowercase `<!doctype html>` (which [is valid HTML](https://html.spec.whatwg.org/multipage/syntax.html#the-doctype)) did not get removed. Now it does by making the regular expression case-insensitive.

This came about because ad markup returned by Amazon APS bids uses lowercase doctype.


patmmccann:
superceded by #246

PR #246, opened 2024-11-18
Replace PostScribe in the PUC

OlenaPostindustria:
Closes [#1018](https://github.com/prebid/prebid-mobile-ios/issues/1018)


src/adHtmlRender.js
```diff
@@ -0,0 +1,11 @@
+export function writeAdHtml(markup, insertHTML = document.body.insertAdjacentHTML.bind(document.body)) {
+    // remove <?xml> and <!doctype> tags
+    // https://github.com/prebid/prebid-universal-creative/issues/134
+    markup = markup.replace(/\<(\?xml|(\!DOCTYPE[^\>\[]+(\[[^\]]+)?))+[^>]+\>/g, '');
```
rickyblaha, line 4: `/g` -> `/gi` to use case-insensitive find & replace. Lowercase doctype is still valid and Amazon APS admarkup uses lowercase doctype. Without that, render of Amazon ads would fail due to the doctype still being present. Feel free to lift the test added over here https://github.com/prebid/prebid-universal-creative/pull/245/files then the other PR could be closed.

patmmccann (approved):

patmmccann (approved):

PR #248, opened 2024-12-20
Mobile app rendering: if burl is missing from bid, write ad markup after MRAID-viewable

rickyblaha:
To reduce discrepancies, when `burl` is missing from mobile app bids, rendering of ad markup is now delayed until MRAID-viewable happens, so tracking pixels present in ad markup do not get called too soon.

Included units tests along with a simple mock of MRAID 2 and MRAID 3 that has only enough features for these tests.

Also included a simple log function called during this rendering sequence to help with debugging. Once ad render already happens, you can use Safari Dev Tools to inspect the web view, go to the console and type `pucLogs` to see all the console logs that happened prior to opening dev tools.


arjunt:
@YuriyVelichkoPI - we should complete the review of this PR and the open measurement impression tracking PR for Mobile.
I understand the there is a plan to release a new version of the PUC, so would be good to complete before that. Thanks.

mmullin:
@ChrisHuie FYI

justadreamer:
discussed in PMC: 
* Yahoo has a powerful feature-flag infrastructure that allows them to conditionally enable such features through a flag passed as a key-value, this allows limited testing
* The suggestion is to integrate a feature-flag support and enable this feature conditionally when a certain key-value is sent
* That should give some confidence in deploying and a way for pubs to opt-in to this feature

PR #274, opened 2025-03-19
Resolve issue where scripts in markup never execute on mobile

OlenaPostindustria:
After merging #246, the scripts inside the markup stopped executing because of `insertAdjacentHTML` behavior. This PR aims to fix this issue.

### DIFF
--- a/test/spec/mobileAndAmpRender_spec.js
+++ b/test/spec/mobileAndAmpRender_spec.js
@@ -287,12 +287,12 @@ describe('writeAdHtml', () => {
   it('removes DOCTYPE from markup', () => {
     const ps = sinon.stub();
     writeAdHtml('<!DOCTYPE html><div>mock-ad</div>', ps);
-    sinon.assert.calledWith(ps, sinon.match.any, '<div>mock-ad</div>')
+    sinon.assert.calledWith(ps, '<div>mock-ad</div>')
   });
 
   it('removes lowercase doctype from markup', () => {
     const ps = sinon.stub();
     writeAdHtml('<!doctype html><div>mock-ad</div>', ps);
-    sinon.assert.calledWith(ps, sinon.match.any, '<div>mock-ad</div>')
+    sinon.assert.calledWith(ps, '<div>mock-ad</div>')
   });
 })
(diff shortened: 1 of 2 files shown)


justadreamer:
This is the root cause of OMID events stopped being sent with PUC 1.17.  Pending an iOS and Android testing report for this fix.

test/spec/mobileAndAmpRender_spec.js
```diff
@@ -287,12 +287,12 @@ describe('writeAdHtml', () => {
   it('removes DOCTYPE from markup', () => {
     const ps = sinon.stub();
     writeAdHtml('<!DOCTYPE html><div>mock-ad</div>', ps);
-    sinon.assert.calledWith(ps, sinon.match.any, '<div>mock-ad</div>')
+    sinon.assert.calledWith(ps, '<div>mock-ad</div>')
```
dgirardi, line 290: Could you add a test case for the regression? "if I writeAdHtml a sript, it should run".

dgirardi:
Are you confident that this fixes https://github.com/prebid/prebid-mobile-ios/issues/1119 (meaning, do the trackers actually fire)? running the unit tests in a webview doesn't really add anything; this was not caused by any difference in webview vs. desktop.

dgirardi (approved):
LGTM as this seems very likely to be the fix, but as mentioned above please verify this against the actual issue if you can - I am unfortunately not set up to easily do this.

OlenaPostindustria:
@dgirardi I was able to see OM signals in Charles after this fix during iOS platform testing.
Charles 4 6 7 - PUC-FixScriptExecution-Testing_iOS 2025-03-19 19-01-05

[PUC-FixScriptExecution-Testing_iOS.chls.zip](https://github.com/user-attachments/files/19347110/PUC-FixScriptExecution-Testing_iOS.chls.zip)

ValentinPostindustria:
Android is also fixed with these changes. 
изображение

justadreamer:
Let's wait a bit with the merge of this please, @ValentinPostindustria would like to run some additional tests on Android.

PR #278, opened 2025-04-22
Restore Postscribe

OlenaPostindustria:
Addresses: https://github.com/prebid/prebid-universal-creative/issues/275

Mobile PUC used `postscribe` for a very good reason - namely because it replaced `document.write` calls that were issued by the loaded creative or its nested scripts.

It turned out not so simple to just dump the creative content into the document via either appending to `innerHtml` or `createContextualFragment` + `document.body.append` (actually using `document.write` kinda works, but it is deprecated, and may cause weird effects), so for now we need to go back to postscribe as it turned out to be essential for mobile, not due to the performance and need for asynchrony, but because of the smart way it dealt with `document.write` calls in the loaded creative.

`document.write` has no effect when called from an async loaded or dynamically added script (which is treated as async loaded), because document is considered loaded and its stream is closed - thus browsers and webviews (Chromium-based and WebKit-based) will not execute `document.write` in such contexts (otherwise it would have to overwrite the whole document with new content).


justadreamer (approved):
Was tested with a scenario similar to described in #279 
- PUC built from the PR branch via `gulp build-dev` producing mobile.js
- Used Proxyman Map Local of cdn.jsdelivr PUC URL to the locally built mobile.js
- Used Proxyman to rewrite prebid cache response and supply 2 test creatives with nested scripts calling `document.write` - the former taken from a real app, the latter is synthesized
- all rendering fine on this PUC dev build: 
Monosnap iPhone 16 2025-05-16 01-22-30

Monosnap iPhone 16 2025-05-16 01-24-31

In the same time without Postscribe (if Map Remote to PUC v 1.17.1) those creatives do not render.  The latter also does not render in a browser with a warning printed to console: `Failed to execute 'write' on 'Document': It isn't possible to write into a document from an asynchronously-loaded external script unless it is explicitly opened.` 

Thus confirming that Postscribe was properly restored in this PR.

PR #287, opened 2025-07-25
Adding support to pass through the account parameter in cookie sync

hhhjort:
Prebid server cookie sync supports receiving the `account` parameter. This change adds support for publishers to pass this parameter through the universal creative cookie sync so that PBS can act upon it. This does not add any functionality using the parameter internally to universal creative code.

### DIFF
--- a/src/cookieSyncWithConsent.js
+++ b/src/cookieSyncWithConsent.js
@@ -285,7 +285,7 @@ function getStringifiedData(endPointArgs) {
         }
     };
     if(BIDDER_ARGS) data.bidders = BIDDER_ARGS;
-    data['account'] = ACCOUNT;
+    if(ACCOUNT) data.account = ACCOUNT;
 
     data = attachConsent(data);
 


src/cookieSync.js
```diff
@@ -24,7 +24,7 @@ const coopSyncParam = parseQueryParam('coop_sync', window.location.search);
 const COOP_SYNC = !coopSyncParam || coopSyncParam === 'true' || !!parseInt(coopSyncParam);
 const GDPR = sanitizeGdpr(parseInt(parseQueryParam('gdpr', window.location.search), 10));
 const GDPR_CONSENT = sanitizeGdprConsent(parseQueryParam('gdpr_consent', window.location.search));
-
+const ACCOUNT = parseQueryParam('account', window.location.search);
```
SyntaxNode, line 27: Nitpick: Please keep the empty line after the `const` section.

src/cookieSync.js
```diff
@@ -267,6 +267,7 @@ function getStringifiedData(endPointArgs) {
     }
   };
   if(BIDDER_ARGS) data.bidders = BIDDER_ARGS;
+  data['account'] = ACCOUNT;
```
SyntaxNode, line 270: I recommend keeping the pattern of `if(ACCOUNT) data.account = ACCOUNT`.

- The `if` statement will avoid setting the parameter to null or empty. 
- The syntax `data['account']` is equivalent to `data.account` and it looks like the latter is preferred here.

^ same for cookieSyncWithConsent.js

PR #290, opened 2025-08-12
Update nativeAssetManager for requestHeightResize fallback

ourcraig:
See https://github.com/prebid/Prebid.js/issues/13749


patmmccann:
fixed in #291

